PR #10, opened 2016-03-02
Fix no query variables exception issue.

dre1080:
Currently the plug doesn't allow one to query without variables and raises the following:

```
        ** (ArgumentError) argument error
            :erlang.iolist_to_binary(%{})
            (poison) lib/poison/parser.ex:35: Poison.Parser.parse/2
            (poison) lib/poison.ex:69: Poison.decode/2
```

This is because the variables passed to be decoded are expected to be `iodata`.

This is to also accomodate for queries as described here: facebook/relay#112

### DIFF
--- a/lib/absinthe/plug.ex
+++ b/lib/absinthe/plug.ex
@@ -54,7 +54,7 @@ defmodule Absinthe.Plug do
     """)
 
     with input when is_binary(input) <- input,
-      {:ok, variables} <- decode_variables(json_codec, variables) do
+      variables <- decode_variables(json_codec, variables) do
         %{variables: variables,
           adapter: config.adapter,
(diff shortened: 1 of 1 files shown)


benwilson512:
Can you show an example of what query is causing this error? None of the test examples https://github.com/absinthe-graphql/absinthe_plug/blob/master/test/lib/absinthe/plug_test.exs#L20 use variables (something I should change) and they all pass.

If we confirm that there is a problem here, please also include a test case in your PR.

Thanks!!

benwilson512:
Ah so I've been able to confirm what the issue is here. When using the content-type: application/json the variables come in already parsed due to plug parsers, and I need to not try to parse them again. I'm gonna do a different branch with fixes I think.

PR #15, opened 2016-03-22
GraphIQL Support

benwilson512:


mix.exs
```diff
@@ -27,6 +28,11 @@ defmodule Absinthe.Plug.Mixfile do
     [applications: [:logger, :plug, :absinthe]]
   end
 
+  defp env(:dev) do
+    [serve_graphiql: true]
```
bruce: Any reason to not just make this `graphiql: true` ?

benwilson512: this is irrelevant now, forgot to delete.

benwilson512:
TODO: I need to make sure `&raw` and accept json is supported. I'm honestly not sure how the latter is different than just doing a regular graphql thing.

bruce:
I've tweaked and merged website#38, waiting to deploy until this is released.

bruce:
Looking at [express-graphql](https://github.com/graphql/express-graphql/blob/master/src/renderGraphiQL.js#L36):

> If you wish to receive JSON, provide the header "Accept: application/json" or add "&raw" to the end of the URL within a browser.

benwilson512:
yeah, the `&raw` bit works, I'm just trying to figure out how doing JSON is different than basically just using the regular plug.

lib/absinthe/plug/graphiql.html.eex
```diff
@@ -0,0 +1,103 @@
...
+    function onEditQuery(newQuery) {
+      parameters.query = newQuery;
+      updateURL();
+    }
+    function onEditVariables(newVariables) {
+      parameters.variables = newVariables;
+      updateURL();
+    }
+    function updateURL() {
+      history.replaceState(null, null, locationQuery(parameters));
```
skosch: Those URLs go into the referrer, and when headers get too long, you run into problems (ninenines/cowboy#939,  graphql/graphiql#159 etc.). Either forgo having the URL store the query, or make sure to send an empty referrer header with the query, if that's possible

PR #23, opened 2016-09-17
Add draft of batch_single_operation

skosch:


bruce:
I think this is great as a pass at this for v1.1. For v1.2, I think we can pretty significantly rework this on top of our Blueprint intermediate representation, which should make a lot of the string-based hacks unnecessary. I'm going to mark this as WIP for now until after v1.2.0 drops and we can spend some time working this out.

skosch:
Sounds good :) When do you expect 1.2 to be ready, roughly?

bruce:
We've pushed out a couple alphas so far; @benwilson512 is tackling the last planned PR and we'll be in beta this week (we've been waiting until we had it in production to do that) -- I would say probably at the end of the week. It depends on the amount/quality of feedback we get from people.

skosch:
Awesome. Looking forward to it!

bruce:
This will need to be rewritten on top of the new v1.2 functionality. Closing this PR, can help with the rewrite in a new PR.

PR #26, opened 2016-10-05
Parse empty operation name as nil

lnikkila:
Rationale from the commit message:

> Some clients e.g. apollo-client [send an empty string][0] instead of `null` as
> the operation name when there is none. The graphql-js reference implementation
> [accepts it][1] as that empty string is falsey. I'm not sure if that's intended
> but since the reference supports it then Absinthe should probably support it as
> well.

Please let me know what you think of this and whether any changes are needed. Thanks!

[0]: https://github.com/apollostack/apollo-client/blob/3ae8b71/src/queries/getFromAST.ts#L52
[1]: https://github.com/graphql/graphql-js/blob/988508c/src/execution/execute.js#L195

### DIFF
--- a/lib/absinthe/plug.ex
+++ b/lib/absinthe/plug.ex
@@ -100,7 +100,7 @@ defmodule Absinthe.Plug do
     """)
 
     variables = Map.get(conn.params, "variables") || "{}"
-    operation_name = conn.params["operationName"]
+    operation_name = conn.params["operationName"] |> decode_operation_name
 
     with {:ok, variables} <- decode_variables(variables, json_codec) do
         absinthe_opts = %{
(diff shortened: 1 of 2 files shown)


benwilson512:
Weird I thought I handled this a while ago. Thanks!

stavro:
@benwilson512 If you can merge this also into the `v1.2` branch, we'd appreciate it!

benwilson512:
done.

beno:
I am running into a (probably) related problem when running Apollo's codegen tool:

    apollo-codegen download-schema http://localhost:8080/graphql --output schema.json

This produces this error:
```
15:29:25.906 [error] #PID<0.475.0> running Contracts.Router terminated
Server: localhost:4000 (http)
Request: POST /graphql
    ** (UndefinedFunctionError) function Plug.Conn.Unfetched.fetch/2 is undefined (Plug.Conn.Unfetched does not implement the Access behaviour)
        (plug) Plug.Conn.Unfetched.fetch(%Plug.Conn.Unfetched{aspect: :params}, "operationName")
        (elixir) lib/access.ex:148: Access.fetch/2
        (elixir) lib/access.ex:179: Access.get/3
        (absinthe_plug) lib/absinthe/plug.ex:126: Absinthe.Plug.prepare/3
        (absinthe_plug) lib/absinthe/plug.ex:89: Absinthe.Plug.execute/2
```
I get this on both version 1.2 and master.

benwilson512:
It looks like your params are not yet fetched. You need to have Plug.Parsers prior to your `plug Absinthe.Plug`.

beno:
Ah yes, slight facepalm there. Thanks!

Michel

PR #33, opened 2016-11-14
Update the locked :absinthe dependency

jparise:
This was previously locked to a checksum that didn't match the 1.2.0
release. Refresh it to point to the (current) 1.2.1 release.

This fixes the build error:
```
** (Mix) Registry checksum mismatch against lock (absinthe 1.2.0)
```


benwilson512:
Thanks!

PR #35, opened 2016-11-23
Add Absinthe.Parser for Plug.Parsers

binaryseed:
@benwilson512 We talked about this idea earlier on Slack. I don't know if you would want to pull this into the library (it'd be a breaking change) but I thought it was interesting what I found while implementing it.

The parser itself is pretty straightforward, and the function `load_body_and_params` inside the Absinthe plug can go away.

I couldn't find anything about `x-www-urlencoded`, but plenty about `x-www-form-urlencoded`, which is handled by the built-in `:urlencoded` parser. With that change in the test setup, and the addition of the `Absinthe.Parser` for GraphQl, we don't need to allow wildcard content type with `pass: ["*/*"]`.

With these changes, the parsers would be configured more like this:

```elixir
plug Plug.Parsers,
  parsers: [:urlencoded, :multipart, :json, Absinthe.Parser],
  json_decoder: Poison
```

### DIFF
--- a/lib/absinthe/plug.ex
+++ b/lib/absinthe/plug.ex
@@ -156,7 +156,7 @@ defmodule Absinthe.Plug do
     end
   end
 
-  def load_body_and_params(%{body_params: %{"query" => query}}=conn) do
+  def load_body_and_params(%{body_params: %{"query" => _}}=conn) do
     {fetch_query_params(conn), ""}
   end
   def load_body_and_params(conn) do
--- a/lib/absinthe/plug.ex
+++ b/lib/absinthe/plug.ex
@@ -156,6 +156,9 @@ defmodule Absinthe.Plug do
     end
   end
 
+  def load_body_and_params(%{body_params: %{"query" => query}}=conn) do
+    {fetch_query_params(conn), ""}
+  end
   def load_body_and_params(conn) do
     case get_req_header(conn, "content-type") do
       ["application/graphql"] ->
--- a/mix.exs
+++ b/mix.exs
@@ -33,7 +33,7 @@ defmodule Absinthe.Plug.Mixfile do
 
   defp deps do
     [
-      {:plug, "~> 1.0"},
+      {:plug, "~> 1.2"},
       {:absinthe, "~> 1.2.0"},
       {:poison, ">= 0.0.0", only: [:dev, :test]},
       {:ex_doc, "~> 0.11.0", only: :dev},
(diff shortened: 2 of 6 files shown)


benwilson512:
This could actually be made backwards compatible if the Plug first checks for a query key in the body, and then if it isn't there it does the existing logic around reading the body and parsing.

binaryseed:
Good point! I've made that change. What do you think about the name of the `Parser`? Maybe `Absinthe.Parser.GraphQl`?

bruce:
Everything in this package should be under `Absinthe.Plug`.

bruce:
And we avoid `GraphQl` or `GraphQL` generally in module naming. Maybe `Absinthe.Plug.Parser` is fine. We certainly want to keep in out of the base `Absinthe` module namespace to avoid confusing with the current parser (and the experimental NIF parser I'm working on as time allows).

binaryseed:
I also just noticed that `Plug.TimeoutError` only exists in `1.2.1` and newer. 

I copied the implementation of the `Parser` on plug `master` - https://github.com/elixir-lang/plug/blob/master/lib/plug/parsers/urlencoded.ex

Should I rev the version of plug to get the latest?

binaryseed:
Slight documentation change here: https://github.com/absinthe-graphql/website/pull/60

lib/absinthe/plug.ex
```diff
@@ -156,6 +156,9 @@ defmodule Absinthe.Plug do
     end
   end
 
+  def load_body_and_params(%{body_params: %{"query" => query}}=conn) do
```
benwilson512: `query` here is unused, and should just be `"query" => _`

benwilson512:
This looks good to go from my perspective, was there any other changes you wanted to make?

binaryseed:
I left a note above about `Plug.TimeoutError` needing `plug` `1.2.1`+ ...

benwilson512:
Well, we should update our mix.exs to require plug 1.2.1 if it's gonna be in here.

binaryseed:
Updated `plug`.. not sure if that's how you'd prefer to specify versions. Let me know

mix.exs
```diff
@@ -33,7 +33,7 @@ defmodule Absinthe.Plug.Mixfile do
 
   defp deps do
     [
-      {:plug, "~> 1.0"},
+      {:plug, ">= 1.2.2"},
```
benwilson512: We probably want just `"~> 1.2"` here

binaryseed:
Updated

binaryseed:
I think this is ready to go!

benwilson512:
Thanks!

PR #36, opened 2016-11-24
Allow custom error in absence of query document

avitex:


benwilson512:
Can you talk about the use case here a little more?

avitex:
Sure, perhaps the API would like to direct a curious user or developer who navigates to the endpoint to some documentation.

So instead of `"No query document supplied"`, it could spit out `"No query document supplied. For information on how to use this API see api.example.com/docs"`, or something similar.

It would only be useful if the GraphQL schema is being hosted on the root "/", as it's the first point a user would hit on their travels.

Perhaps I'm missing something? Is their a fall-though ability?

benwilson512:
Makes sense, thanks for the PR!

I mostly just want to avoid extending this into arbitrarily complex handling around this. Somewhere along the line it makes more sense to just do a custom plug. BUT this is nice, simple, and I think covers a lot of use cases. Good stuff, thanks!

avitex:
No problems, thanks again for the terrific libraries!

PR #39, opened 2016-12-13
Batching plug

skosch:
Rough implementation based on @benwilson512's ideas.

### DIFF
--- a/lib/absinthe/plug/batch.ex
+++ b/lib/absinthe/plug/batch.ex
@@ -1,0 +1,284 @@
+defmodule Absinthe.Plug.Batch do
+  @moduledoc """
+  A plug for using Absinthe
+
+  See [The Guides](http://absinthe-graphql.org/guides/plug-phoenix/) for usage details
+  """
+
+  @behaviour Plug
+  import Plug.Conn
+  require Logger
+
+  alias Absinthe.Phase
+
+  @type function_name :: atom
+
+  @type opts :: [
+    schema: atom,
+    adapter: atom,
+    path: binary,
+    context: map,
+    json_codec: atom | {atom, Keyword.t},
+    no_query_message: binary,
+  ]
+
+  @doc """
+  Sets up and validates the Absinthe schema
+  """
+  @spec init(opts :: opts) :: map
+  def init(opts) do
+    adapter = Keyword.get(opts, :adapter)
+    context = Keyword.get(opts, :context, %{})
+    
+    no_query_message = Keyword.get(opts, :no_query_message, "No query document supplied")
+    wrong_http_method_message = Keyword.get(opts, :wrong_http_method_message, "Can only perform batch queries from a POST request")
+
+    json_codec = case Keyword.get(opts, :json_codec, Poison) do
+      module when is_atom(module) -> %{module: module, opts: []}
+      other -> other
+    end
+
+    # This function is applied to each result tuple in the shape of {%{data: ...}, "1"}
+    # before sending the result down the wire. As a default, we'll use 
+    # the format required by react-relay-network-layer's batching mechanism
+    payload_formatter = Keyword.get(opts, :payload_formatter, {__MODULE__, :default_payload_formatter})
+
+    schema_mod = opts |> get_schema
+
+    %{adapter: adapter, schema_mod: schema_mod, context: context, json_codec: json_codec,
+      payload_formatter: payload_formatter, no_query_message: no_query_message,
+      wrong_http_method_message: wrong_http_method_message
+    }
+  end
+
+  defp get_schema(opts) do
+    default = Application.get_env(:absinthe, :schema)
+    schema = Keyword.get(opts, :schema, default)
+    try do
+      Absinthe.Schema.types(schema)
+    rescue
+      UndefinedFunctionError ->
+        raise ArgumentError, "The supplied schema: #{inspect schema} is not a valid Absinthe Schema"
+    end
+    schema
+  end
+
+  @doc """
+  Parses, validates, resolves, and executes the given Graphql Document
+  """
+  def call(conn, %{json_codec: json_codec} = config) do
+    {conn, result} = conn |> execute(config)
+
+    case result do
+      {:input_error, msg} ->
+        conn
+        |> send_resp(400, msg)
+
+      {:ok, result} ->
+        conn
+        |> json(200, result, json_codec)
+
+      {:error, {:http_method, text}, _} ->
+        conn
+        |> send_resp(405, text)
+
+      {:error, error, _} when is_binary(error) ->
+        conn
+        |> send_resp(500, error)
+
+    end
+  end
+
+  @doc false
+  def execute(conn, config) do
+    {conn, body} = load_body_and_params(conn)
+
+    with {:ok, body} <- ensure_http_post_method(conn, body, config),
+    {:ok, prepared_queries} <- prepare_queries(conn, body, config),
+    {:ok, queries} <- validate_query_list(prepared_queries, config.no_query_message) do
+
+      prepared_documents = prepare_documents(conn, config, queries)
+
+      # Check if any of the prepared documents have errors. At this stage,
+      # this could mean they have non-empty :error fields *OR* they are simply
+      # an {:input_error, msg}.
+
+      {valid_prepared_documents_with_ids, invalid_documents_with_ids} = queries
+        |> Enum.map(&Keyword.get(&1.opts, :query_id, nil))
+        |> Enum.zip(prepared_documents)
+        |> Enum.partition(fn
+          {_query_id, %{errors: errors}} -> Enum.empty?(errors) # if true, this is valid
+          {_query_id, {:input_error, _}} -> false
+          {_query_id, {:error, _}} -> false
+        end)
+
+      valid_prepared_documents = valid_prepared_documents_with_ids  
+        |> Enum.map(fn {_, doc} -> doc end)
+
+      resolved_documents = resolve_documents(conn, config, valid_prepared_documents)
+      error_documents = format_invalid_documents(conn, config, invalid_documents_with_ids)
+
+      {conn, {:ok, resolved_documents ++ error_documents}}
+    else
+      {:input_error, msg} -> {conn, {:input_error, msg}}
+      {:error, msg} -> {conn, {:error, msg}}
+    end
+  end
+  
+  def prepare_documents(conn, config, queries) do
+    queries
+    |> Enum.map(fn q ->
+      with preparation_pipeline <- setup_pipeline(conn, config, {__MODULE__, :preparation_pipeline}, q.opts),
+      {:ok, prepared_document, _} <- Absinthe.Pipeline.run(q.query_string, preparation_pipeline) do
+        prepared_document
+      else
+        {:error, msg} -> {:error, msg}
+        msg -> {:input_error, msg}
+      end
+    end)
+  end
+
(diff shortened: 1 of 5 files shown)


lib/absinthe/plug/batch.ex
```diff
@@ -0,0 +1,314 @@
...
+      query_id: query_id,
+      jump_phases: false,
+    ]
+    %{query_string: query_string, opts: absinthe_opts}
+  end
+
+  defp validate_query_list(queries, no_query_message) do
+    all_queries_valid = queries
+      |> Enum.all?(fn q ->
+        String.valid?(q.query_string)
```
skosch: I vaguely remember encountering a case where `query_string` was nil because the input had a syntax error and couldn't be parsed properly. I suppose `not is_nil` would have the same effect? Or we could just trust that Relay/Apollo wouldn't send mangled query arrays?

lib/absinthe/plug/batch_phases/batch_resolution.ex
```diff
@@ -0,0 +1,88 @@
...
+    new_blueprint = %Absinthe.Blueprint{
+      schema: hd(blueprints).schema,
+      errors: Enum.reduce(blueprints, [], fn (bp, e) -> bp.errors ++ e end),
+      operations: [
+        %Absinthe.Blueprint.Document.Operation{
+          current: true,
+          fields: all_fields,
+          name: "__merged_batch_operation",
+          type: hd(hd(blueprints).operations).type,
+          schema_node: hd(hd(blueprints).operations).schema_node,
```
skosch: yes please.

lib/absinthe/plug/batch.ex
benwilson512: Ah. Yeah we should explicitly ignore nil, which really just means we can use `all?/1` because `nil` evaluates as false.

lib/absinthe/plug/batch_phases/batch_resolution.ex
```diff
@@ -0,0 +1,88 @@
...
+    operation = operations
+      |> Enum.find(&(&1.current))
+
+    fields = for field <- operation.fields do
+      Absinthe.Blueprint.put_flag(field, :query_id, query_id)
+    end
+    fields
+  end
+
+  defp get_emitter_from_results(results) do
```
skosch: Do you mean the Map.get or the unless/else?

lib/absinthe/plug/batch.ex
skosch: Ah, I remember now ... this was a short-circuit thing where if `q.query_string` didn't exist, the comparison `q.query_string != ""` would crash. Since `query_string` is always present now, I agree that `all?/1` should do the job. Although it may still be a good idea to make sure that `query_id` also isn't nil.

lib/absinthe/plug/batch_phases/batch_resolution.ex
benwilson512: Well in general I'm pretty sure this function could just be:
```
def get_emitter_from_results([%{emitter: emitter} | _]), do: emitter
def get_emitter_from_results(_), do: nil
```

However now that I look at it further I'm not sure that the place calling this function is doing the right thing.

lib/absinthe/plug/batch.ex
```diff
@@ -0,0 +1,314 @@
...
+  def setup_pipeline(conn, config, pipeline, opts) do
+    private = conn.private[:absinthe] || %{}
+    private = Map.put(private, :http_method, conn.method)
+    config = Map.put(config, :conn_private, private)
+
+    {module, fun} = pipeline
+    apply(module, fun, [config, opts])
+  end
+
+  def preparation_pipeline(config, opts) do
```
skosch: I'd like for it to be `defp`, except `apply` (on line 267) won't work with private functions. Is there a workaround to that?

benwilson512: Ah so in general I was hoping we'd be able to get rid of a lot of the `apply`s because they make following the code a fair bit harder. You have to back track to figure out what the module and function are, and often they're literals (ie, not user supplied). I only see `apply` being necessary if for some reason it was some user supplied module / function pair.

skosch: :+1: I'm not married to `apply` at all. It's what was used in the vanilla plug I started with so I just went along with it. Let's call those functions directly then and make them private.

lib/absinthe/plug/batch_phases/batch_resolution.ex
```diff
@@ -0,0 +1,88 @@
...
+
+  @spec run([Blueprint.t], Keyword.t) :: Phase.result_t
+  def run(blueprints, options \\ []) do
+
+    all_fields = blueprints
+      |> get_all_fields
+
+    new_blueprint = %Absinthe.Blueprint{
+      schema: hd(blueprints).schema,
+      errors: Enum.reduce(blueprints, [], fn (bp, e) -> bp.errors ++ e end),
```
skosch: That's a good question. I remember putting this in after not seeing an error I wanted to see, but if it's still needed I can't remember right now. Hm ...

lib/absinthe/plug/batch_phases/batch_resolution.ex
```diff
@@ -0,0 +1,88 @@
...
+
+    {:ok, results}
+  end
+
+  defp get_all_fields(blueprints) do
+    blueprints
+    |> Enum.map(&get_fields_with_query_id/1)
+    |> Enum.reduce([], fn (fields, all_fields) ->
+      fields ++ all_fields
+    end)
```
benwilson512: Pretty sure if you did `Enum.flat_map` you could remove the `Enum.reduce` call.

lib/absinthe/plug/batch_phases/batch_resolution.ex
skosch: also, the way I'm creating this new operation, is that generally kosher? Is there anything missing or redundant? The `type`, for example; does it make sense to just hardcode/require that to be a query (since I don't think mutations will ever be batched)?

lib/absinthe/plug/batch_phases/batch_resolution.ex
```diff
@@ -0,0 +1,88 @@
...
+  end
+
+  defp get_fields_with_query_id(%{flags: %{query_id: query_id}, operations: operations}) do
+    operation = operations
+      |> Enum.find(&(&1.current))
+
+    fields = for field <- operation.fields do
+      Absinthe.Blueprint.put_flag(field, :query_id, query_id)
+    end
+    fields
```
benwilson512: The reassignment is redundant, just have the `for` comprehension be the last thing in the function.

I'm a bit confused though, why is it called `get_fields_with_query_id` if it seems to be setting a flag and not filtering by existing flags.

benwilson512:
- [ ] rework how `put_flag` is used (don't worry about this for now, this is just a note so I don't forget to mention this later).

lib/absinthe/plug/batch_phases/batch_resolution.ex
skosch: Good catch.

lib/absinthe/plug/batch_phases/batch_resolution.ex
benwilson512: I think it's fine as is. You do bring up a good point though about only batching queries, we need to do a filter on the submitted documents to make sure that they're only queries.

lib/absinthe/plug/batch_phases/batch_resolution.ex
skosch: Yep, you're right on both counts.

benwilson512: On the latter note, is the function named badly or is it not doing the filtering that it should be doing?

skosch: Just a confusing name. It's supposed to a) grab the current operation, and b) tag the fields with the query id.
Really it'd be neater for the function to only do the latter, and we could call it something sensible like `put_query_id_flags`.

benwilson512: Where does that fit in with https://github.com/absinthe-graphql/absinthe_plug/pull/39/files/f820eeb6ad53d690e5b5797d10aff510cdc0509c#diff-375404e7f6c0ec9cae60ae9e8b5c991cR11 which seems to do the same thing?

skosch: Not sure what line that link is pointing to. Do you mean the other `put_flag`? That tags the blueprint, whereas this tags the field.

benwilson512: Ah yes of course, missed that.

lib/absinthe/plug/batch_phases/batch_resolution.ex
skosch: :+1: on the function pattern match. What is it you find problematic?

lib/absinthe/plug/batch_phases/batch_resolution.ex
skosch: Although in hindsight that really is confusing. There should be comments there

lib/absinthe/plug/batch_phases/batch_resolution.ex
benwilson512: Well at the end of the day the emitter for the operation result should be the operation blueprint. One way to get that is from the operation results (as you're doing) but presumably we already have the operation blueprint around earlier. This means we can always have a blueprint whether or not the operation returns results.

skosch: So we can just use `blueprint` for `emitter`?

lib/absinthe/plug/batch_phases/batch_resolution.ex
```diff
@@ -0,0 +1,88 @@
...
+
+    results = for blueprint <- blueprints do
+        query_id = blueprint.flags.query_id
+        this_operations_results = Enum.filter(result_data, fn data ->
+          data.emitter.flags.query_id == query_id
+        end)
+
+        result_object = %Absinthe.Blueprint.Document.Resolution.Object{
+          fields: this_operations_results,
+          emitter: get_emitter_from_results(this_operations_results),
```
benwilson512: To move the discussion up here: is `blueprint` here the operation blueprint? quite frankly I've sort of lost track. I generally try to only use a variable named `blueprint` when referring to top level `%Blueprint{}` structs. Otherwise I use `operation` or `operation_blueprint` to be clearer.

skosch: Yeah my `operation`/`blueprint` naming is problematic. Because only the "current" operation is executed for each blueprint, I ended up using them somewhat interchangeably and that's confusing.

Here, `blueprints` is the list of all top-level parsed-and-validated blueprints, from before they were merged into `new_blueprint` and executed as one. After having resolved, we loop through them and create a `Resolution.Object` for each one, filling it with `this_operations_results` (i.e. those result fields matching the `blueprint`'s query id). Indeed, `this_operations_results` should probably be called `results_for_current_blueprint` or something.

Does that help?

skosch: How about a rename like
`blueprints` -> `all_blueprints`
`blueprint` -> `current_blueprint`
`result_data` -> `all_result_fields`
`query_id` -> `current_query_id`
`this_operations_results` -> `results_for_current_blueprint`
`results` -> `resolution_objects`

PR #43, opened 2017-01-06
Fancy batching plug

benwilson512:


benwilson512:
- [ ] Handle customization of pipelines
- [ ] refactor the various FIXME tags.
- [ ] check on whether `{:error` tuples need to be handled in inner `def execute` stuff.

skosch:
I just noticed Apollo doesn't let you specify a separate URL for the batching endpoint, so it would be nice if `absinthe_plug` recognized incoming JSON arrays as batched queries :)

bruce:
Is this in a good place, testing wise, where I can build a PR that would re-sit the processing here on top of the refactoring I've done in #53, and still make sure everything works? Or is it non-functional?

benwilson512:
It largely works but a lot of the hackyness where will get wildly easier after the changes related to middleware goes thorugh. I'd like to wait until that/s gone

Issue #44, opened 2017-01-07
Fails to load body when the Content-Type encoding is set 


ScrimpyCat:
When receiving a GraphQL query document (`Content-Type: application/graphql`) if the the encoding is also passed (or any other parameters for that matter), the plug will assume it's not a GraphQL document.

e.g. `Content-Type: application/graphql; charset=UTF-8` will not work.

The issue is because of this line https://github.com/absinthe-graphql/absinthe_plug/blob/master/lib/absinthe/plug.ex#L169 the match here is matching exactly the string `"application/graphql"`. So any other combination of parameters or encodings will obviously not match and fall through to the empty body case (`{conn, ""}`).

A simple fix is to just match `[<<"application/graphql", _ :: binary>>]` instead (though this is most likely wrong as any result as long as it contains that prefix will pass). However I'm unsure what the appropriate fix should be. I assume the GraphQL spec states the encoding should implicitly be UTF-8 (like JSON). So maybe the correct scenario should be a `Content-Type` whose parameters are either `application/graphql` only, or `application/graphql` with an explicit encoding set to UTF-8. While any other parameters/combination of (or other encodings) would be a failure.

I'm happy to submit a fix, just wondering what the ideal solution should cover.

ScrimpyCat:
I just went about doing a fix, but noticed the tests would pass regardless. Wasn't sure why at first, but it looks like it's because `Absinthe.Plug.Parser` is now being used to handle `application/graphql` documents and so a different path is taken (specifically https://github.com/absinthe-graphql/absinthe_plug/blob/master/lib/absinthe/plug.ex#L164).

So I'm wondering in the next absinthe_plug version, will we be expected to make sure we add the `Absinthe.Plug.Parser` to our plug parsers? Or will it be optional?

If it's optional then the scenario mentioned in the issue will still be a problem (will need fixing). If it'll be required, couldn't the `["application/graphql"]` match clause be removed? Since if the `Absinthe.Plug.Parser` is used then no `application/graphql` will fall through to that function. It could then become something like:

```elixir
def load_body_and_params(%{body_params: %{"query" => _}}=conn) do
  {fetch_query_params(conn), ""}
end
def load_body_and_params(conn), do: {conn, ""}
```

binaryseed:
`Absinthe.Plug.Parser` was added here: https://github.com/absinthe-graphql/absinthe_plug/pull/35

It was introduced as an optional parser, so that it would be backward compatible.

[...]

PR #47, opened 2017-01-22
Allow application/graphql with encoding

ScrimpyCat:
Fix for the bug discussed in https://github.com/absinthe-graphql/absinthe_plug/issues/44


benwilson512:
Per my note on that issue, I think the best route is to indicate that using the plug parser is mandatory.

PR #48, opened 2017-01-25
Add no-referrer header

skosch:
No-referrer header to prevent long queries stored in the URL bar `location` from being sent as referrer headers. Long headers can cause cowboy to silently drop the request.


benwilson512:
Cool so this solves the issue you had?

skosch:
No more problems so far!

PR #52, opened 2017-02-13
Custom adapter not passed through to absinthe pipeline/run function

russmatney:
Related to [the adapter fix made here](https://github.com/absinthe-graphql/absinthe/pull/265), adapters are not being passed through properly via:

```
  plug Absinthe.Plug,
    adapter: SuperSavvyLanguageAdapter,
    schema: Web.GraphQL.Schema
```

### DIFF
--- a/lib/absinthe/plug.ex
+++ b/lib/absinthe/plug.ex
@@ -193,6 +193,12 @@ defmodule Absinthe.Plug do
           root_value: (conn.private[:absinthe][:root_value] || %{}),
           operation_name: operation_name,
         ]
+        absinthe_opts =
+          if not(is_nil(config.adapter)) do
+            Keyword.put(absinthe_opts, :adapter, config.adapter)
+          else
+            absinthe_opts
+          end
         {:ok, raw_input, absinthe_opts}
     end
   end


lib/absinthe/plug.ex
```diff
@@ -193,6 +193,12 @@ defmodule Absinthe.Plug do
           root_value: (conn.private[:absinthe][:root_value] || %{}),
           operation_name: operation_name,
         ]
+        absinthe_opts =
+          if not(is_nil(config.adapter)) do
+            Keyword.put(absinthe_opts, :adapter, config.adapter)
+          else
+            absinthe_opts
+          end
```
benwilson512: This logic should happen in the `init` portion of the plug.

bruce:
Note that I've fixed this separately for the v1.3.0 release in #52. (But if we want to fix this is a v1.2.X release, continue on.)

russmatney:
Thanks for the heads up @bruce - I won't duplicate the effort or introduce any conflicts, glad to see it'll be fixed going forward. For now we're using the [PR referenced above](https://github.com/absinthe-graphql/absinthe/pull/265) internally to ensure our adapter is used. Will wait to see what implementation decision is there.

PR #53, opened 2017-02-16
Document Providers (supporting "Persisted Queries")

bruce:
This PR introduces changes slated for the next minor release, v1.3.0.

## Document Providers

This PR ads the concept of "document providers," which constitute a flexible mechanism for defining how GraphQL requests are processed (how GraphQL documents are "found," and how to apply Absinthe pipelines to execute them).

There are currently two types of document provider defined:

- `Absinthe.Plug.DocumentProvider.Default`, which is the behavior we're already used to -- GraphQL documents are provided directly by clients, pulled from the body/params, and the full (possibly custom-configured) pipeline is used to process them.
- `Absinthe.Plug.DocumentProvider.Compiled`, which can be used to build document providers that "compile in" partially processed documents (that are stored by key), and then only have to process documents through the remaining portion of the pipeline when requests (for specific document keys) come in.

These document providers can be used together (allowing both ad hoc and predetermined documents), or separately (eg, disallowing ad hoc queries).

### Compiled Document Providers

Using this second document provider type, `Absinthe.Plug.DocumentProvider.Compiled`, it's trivial to add support for Apollo's "Persisted Queries" (for background, read [the blog post](https://dev-blog.apollodata.com/persisted-graphql-queries-with-apollo-client-119fd7e6bba5#.ivl34kll7)) by simply creating a document provider module and building in the documents that Apollo's `persistedgraphql` tool stores in `extracted_queries.json`:

```elixir
defmodule MyApp.ExtractedQueryProvider do
  use Absinthe.Plug.DocumentProvider.Compiled

  provide File.read!("/path/to/extracted_queries.json")
  |> Poison.decode!
  |> Map.new(fn {k, v} -> {v, k} end) # invert key/value
end
```

Then, configuring your plug to support the document type:

```elixir
plug Absinthe.Plug,
  schema: MyApp.Schema,
  document_providers: [
    Absinthe.Plug.DocumentProvider.Default,
    MyApp.ExtractedQueryProvider
  ]
```

Then, any request whose`"id"` param matches a document will have that document processed --without having to repay the cost for parsing, Absinthe.Blueprint conversion, and many structural validations.

### Document Provider Extensibility

The document provider pattern could be used in the future to build more flexible, dynamic options (e.g., a smart cache that stores common queries for some TTL). The `Absinthe.Plug.DocumentProvider` behaviour has a simple contract.

## Refactoring

A sizable amount of logic inside `Absinthe.Plug` responsible for extracting various values for use in processing has been extracted into a new module/struct, `Absinthe.Plug.Request`, which is used by both `Absinthe.Plug` and `Absinthe.Plug.GraphiQL`.

## Documentation

A sizable portion of the package has had documentation and spec information added.


bruce:
Notably, this removed document logging entirely. While this needs to be re-added, since this is targeting the `next` branch for the v1.3.0 release, let's do that in a different PR. We've needed better control of logging for some time. I'll look at it tomorrow.

README.md
```diff
@@ -37,27 +38,54 @@ def deps do
...
+
+forward "/api", Absinthe.Plug,
+  schema: MyApp.Schema
+
+For more information, see the API documentation for `Absinthe.Plug`.
+
+## GraphiQL
 
-### Roadmap
+To add support for a GraphiQL interface, add a configuration for
```
jparise: Might be nice to link to https://github.com/graphql/graphiql here.

lib/absinthe/plug.ex
```diff
@@ -1,92 +1,93 @@
...
   import Plug.Conn
   require Logger
 
-  @type function_name :: atom
-
+  @typedoc """
+  - `:adapter` -- (Optional) Absinthe adapter to use (default: `Absinthe.Adapter.LanguageConventions`).
+  - `:context` -- (Optional) Initial value for the Absinthe context, available to resolvers. (default: `%{}`).
+  - `:no_query_message` -- (Optional) Message to return to the client if no query is provided (default: "No query document supplied").
+  - `:json_codec` -- (Optional) JSON codec to use (default: `Poison`).
```
jparise: Document the `{atom, Keyword.t}` value format?

lib/absinthe/plug.ex
```diff
@@ -1,92 +1,93 @@
...
+  - `:pipeline` -- (Optional) `{module, atom}` reference to a 2-arity function that will be called to generate the processing pipeline. (default: `{Absinthe.Plug, :default_pipeline}`).
+  - `:document_providers` -- (Optional) A `{module, atom}` reference to a 1-arity function that will be called to determine the document providers that will be used to process the request. (default: `{Absinthe.Plug, :default_document_providers}`, which configures `Absinthe.Plug.DocumentProvider.Default` as the lone document provider). A simple list of document providers can also be given. See `Absinthe.Plug.DocumentProvider` for more information about document providers, their role in procesing requests, and how you can define and configure your own.
+  - `:schema` -- (Required, if not handled by Mix.Config) The Absinthe schema to use. If a module name is not provided, `Application.get_env(:absinthe, :schema)` will be attempt to find one.
+  """
   @type opts :: [
```
jparise: I think some of these could use `module` instead of `atom` if we want to be very specific.

lib/absinthe/plug.ex
```diff
@@ -1,92 +1,93 @@
...
-
+  @typedoc """
+  - `:adapter` -- (Optional) Absinthe adapter to use (default: `Absinthe.Adapter.LanguageConventions`).
+  - `:context` -- (Optional) Initial value for the Absinthe context, available to resolvers. (default: `%{}`).
+  - `:no_query_message` -- (Optional) Message to return to the client if no query is provided (default: "No query document supplied").
+  - `:json_codec` -- (Optional) JSON codec to use (default: `Poison`).
+  - `:pipeline` -- (Optional) `{module, atom}` reference to a 2-arity function that will be called to generate the processing pipeline. (default: `{Absinthe.Plug, :default_pipeline}`).
+  - `:document_providers` -- (Optional) A `{module, atom}` reference to a 1-arity function that will be called to determine the document providers that will be used to process the request. (default: `{Absinthe.Plug, :default_document_providers}`, which configures `Absinthe.Plug.DocumentProvider.Default` as the lone document provider). A simple list of document providers can also be given. See `Absinthe.Plug.DocumentProvider` for more information about document providers, their role in procesing requests, and how you can define and configure your own.
```
jparise: I don't think this function's argument (map) is documented here or in the `DocumentProviders` documentation.

lib/absinthe/plug/document_provider/compiled/check.ex
```diff
@@ -0,0 +1,51 @@
...
+      module,
+      Enum.map(errors, &format_error/1)
+    )
+  end
+
+  def format_error(%{locations: [%{line: line}|_], message: message, phase: phase}) do
+    "On line #{line}: #{message} (#{phase})"
+  end
+  def format_error(error) do
+    "#{inspect(error)}"
```
jparise: Could this be just `inspect(error)` (without the wrapping string)?

lib/absinthe/plug/graphiql.ex
```diff
@@ -34,16 +47,15 @@ defmodule Absinthe.Plug.GraphiQL do
     interface: atom
```
jparise: Could be a separate change, but `:simple | :advanced`?

test/lib/absinthe/plug/document_provider/compiled_test.exs
```diff
@@ -0,0 +1,89 @@
...
+  test ".get compiled" do
+    assert %Absinthe.Blueprint{} = Compiled.get(LiteralDocuments, "1")
+    assert %Absinthe.Blueprint{} = Compiled.get(LiteralDocuments, "1", :compiled)
+  end
+
+  test ".get source" do
+    assert @foo_query == Compiled.get(LiteralDocuments, "1", :source)
+  end
+
+  def literal(_) do
```
jparise: Are these two functions (`literal/1` and `extracted/1`) used?

README.md
bruce: Maybe -- our default isn't the standard GraphiQL interface, however (as it's seriously lacking), so I think in the end we'll have separate links to https://github.com/graphql/graphiql and https://github.com/OlegIlyenko/graphiql-workspace (once that gets updated, hopefully in the next week).

lib/absinthe/plug.ex
bruce: Agreed.

lib/absinthe/plug/document_provider.ex
```diff
@@ -0,0 +1,111 @@
...
+  the document providers that are defined in this package for more information.
+
+  - `Absinthe.Plug.DocumentProvider.Default`
+  - `Absinthe.Plug.DocumentProvider.Compiled`
+  """
+
+  @typedoc """
+  The configuration for a document provider (when it needs options)
+  """
+  @type with_options :: {module, Keyword.t}
```
fishcakez: I think it would be clearer to inline this type in `t`, name is awkward and indirection is unhelpful to users. For private specs can use `{module, Keyword.t}`.

lib/absinthe/plug/request.ex
```diff
@@ -0,0 +1,217 @@
...
+  defp extract_root_value(conn) do
+    conn.private[:absinthe][:root_value] || %{}
+  end
+
+  #
+  # DOCUMENT PROVIDERS
+  #
+
+  @spec calculate_document_providers(map) :: [Absinthe.Plug.DocumentProvider.t]
+  defp calculate_document_providers(%{document_providers: {module, fun}} = config) do
```
fishcakez: We should have guard `when is_atom(fun)` otherwise we can't have a single document provider of that is a `Absinthe.Plug.Document.Provider.t` of the form `{module, Keyword.t}`. If this isn't supported then we don't need `List.wrap` call, the docs suggest for `document_providers` say its not.

lib/absinthe/plug.ex
```diff
@@ -1,92 +1,93 @@
...
+    schema: module,
+    adapter: module,
     context: map,
-    json_codec: atom | {atom, Keyword.t},
-    pipeline: {Module.t, function_name},
-    no_query_message: binary,
+    json_codec: module | {module, Keyword.t},
+    pipeline: {module, atom},
+    no_query_message: String.t,
+    document_providers: [Absinthe.Plug.DocumentProvider.t]
```
fishcakez: I think this should be a non empty list of providers, a single provider (not doc'ed but List.wrap is used later) or a module/function:
```elixir
document_providers: [Absinthe.Plug.DocumentProvider.t, ...] | Absinthe.Plug.DocumentProver.t | {module, atom}
```

lib/absinthe/plug/request.ex
```diff
@@ -0,0 +1,217 @@
...
+    apply(module, fun, [config])
+  end
+  defp calculate_document_providers(%{document_providers: simple_value}) do
+    List.wrap(simple_value)
+  end
+
+  @spec ensure_document_providers!([Absinthe.Plug.DocumentProvider.t]) :: [Absinthe.Plug.DocumentProvider.t] | no_return
```
fishcakez: Should only spec `no_return` when there is never a return because it prevents dialyzer from warning (with default options) when only no return occurs. Also the empty list case is an error so shouldn't be spec'ed. Rather it should be something like:
```elixir
@spec ensure_document_providers!(providers) :: providers when providers: [Absinthe.Plug.DocumentProvider.t, ...]
```

test/lib/absinthe/plug/document_provider/compiled_test.exs
fishcakez: These are called as `document_providers`, see `document_providers: {__MODULE__, :literal}` etc

lib/absinthe/plug.ex
```diff
@@ -1,92 +1,93 @@
...
+  - `:json_codec` -- (Optional) A `module` or `{module, Keyword.t}` dictating which JSON codec should be used (default: `Poison`).
+  - `:pipeline` -- (Optional) `{module, atom}` reference to a 2-arity function that will be called to generate the processing pipeline. (default: `{Absinthe.Plug, :default_pipeline}`).
+  - `:document_providers` -- (Optional) A `{module, atom}` reference to a 1-arity function that will be called to determine the document providers that will be used to process the request. (default: `{Absinthe.Plug, :default_document_providers}`, which configures `Absinthe.Plug.DocumentProvider.Default` as the lone document provider). A simple list of document providers can also be given. See `Absinthe.Plug.DocumentProvider` for more information about document providers, their role in procesing requests, and how you can define and configure your own.
+  - `:schema` -- (Required, if not handled by Mix.Config) The Absinthe schema to use. If a module name is not provided, `Application.get_env(:absinthe, :schema)` will be attempt to find one.
```
fishcakez: Using Mix.Config to set a default config where there can be multiple orthogonal configurations is usually bad practise. A change for one configuration could affect another.

bruce: This is longstanding behavior and a change wouldn't really in the scope of this PR -- happy to discuss it as part of a different one.

lib/absinthe/plug/document_provider/compiled/check.ex
bruce: @jparise There used to be indentation; the wrapping string is an artifact of that old approach. Thanks, I'll remove it.

@fishcakez I'm not sure which `handle_node/2` you're referencing; with `jump_phases: false`, `Absinthe.Phase.Parse` can return a value that makes `error` here just a `String.t` -- and ends the pipeline early. I'm not sure, but perhaps a spec is off.

(I'd rather Phase results always returned error structs, but there are a couple cases where plain strings are returned -- mostly for backwards compatibility with tests we used to check 1.1 -> 1.2, etc). It's certainly something I'd like to clean up.

lib/absinthe/plug/graphiql.ex
bruce: Thanks; probably `nil | :advanced | :simple`, as it's an optional value.

lib/absinthe/plug/request.ex
bruce: Good catch, thanks.

lib/absinthe/plug/request.ex
bruce: Thanks; this is very useful. (My use of specs is pretty basic, I'll admit; I wasn't familiar with `when` in this context.)

test/lib/absinthe/plug/document_provider/compiled_test.exs
bruce: @fishcakez is correct, but I'll simplify these -- the use here is supported, but dates before I supported passing the document provider values directly as the option.

lib/absinthe/plug/document_provider/compiled/check.ex
fishcakez: The spec is at https://github.com/absinthe-graphql/absinthe_plug/pull/53/files/49fb8e7c9bff060d1e1632dfb100bfe90431bfee#diff-894bffdabba6a760f486a15b84332d6fR28, which means `errors` must be `[Phase.Error.t]`, and so `format_error` is passed an `Phase.Error.t`

lib/absinthe/plug/graphiql.ex
fishcakez: The default is `:advanced` if it is not present. `interface: nil` would crash with `CaseClauseErrr`: https://github.com/absinthe-graphql/absinthe_plug/pull/53/files/c5ad9a06ecb32e10e548bc268d3850f40d9657de#diff-c40bc49b61fa68956d5b9309edae59a4R93.

bruce:
Merging this into our branch for v1.3 (`next`) and cutting an alpha release.

PR #58, opened 2017-02-24
Plug configuration option to disable the default query request logging.

dinsley:
The main use-case that we have for this feature is that we work in the healthcare industry and this logging needs to be disabled so that we can prevent personal health information (PHI) from leaking into our logs. We're logging from outside the scope of this Plug and scrubbing the data ourselves.

The option defaults to `true` so that the current behaviour is maintained. 

Tests are missing from this PR as I was unable to get `ExUnit.CaptureLog` or `ExUnit.CaptureIO` functioning properly, so that I could write tests for the existing logging behaviour before the change and the new behaviour with the new option.

An Example configuration to disable the logging from within `web.ex` in a Phoenix project:

```  
forward "/", Absinthe.Plug,
    schema: Example.Schema,
    log: false
```

### DIFF
--- a/lib/absinthe/plug.ex
+++ b/lib/absinthe/plug.ex
@@ -77,7 +77,8 @@ defmodule Absinthe.Plug do
     pipeline: {Module.t, function_name},
     no_query_message: binary,
     analyze_complexity: boolean,
-    max_complexity: non_neg_integer | :infinity
+    max_complexity: non_neg_integer | :infinity,
+    log: boolean
   ]
 
   @doc """
@@ -98,11 +99,14 @@ defmodule Absinthe.Plug do
     end
 
     schema_mod = opts |> get_schema
+    adapter = Keyword.get(opts, :adapter)
+
+    log = Keyword.get(opts, :log, true)
 
     raw_options = Keyword.take(opts, @raw_options)
 
     %{adapter: adapter, schema_mod: schema_mod, context: context, json_codec: json_codec,
-      pipeline: pipeline, no_query_message: no_query_message,
+      pipeline: pipeline, no_query_message: no_query_message, log: log,
       raw_options: raw_options}
   end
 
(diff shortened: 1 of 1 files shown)


bruce:
I'd like this to serve as the starting point for a wider discussion of how logging options should be handled, especially as logging support hasn't been added yet in the refactoring being done in #53 (slated for v1.3).

While turning off logging entirely should certainly be supported, I'd like to make sure we also handle these cases, at a minimum:

- The format of the log. Can we offer a condensed view?
- The inclusion (or scrubbing) of variables (in some use cases this may be enough to support the core reason for this PR, and wider concerns, like passwords)
- Supporting logging only certain documents, or logging them at different levels
- Supporting logging results/statistics (or perhaps that's left to related phases)
- Configuring logging via config

Perhaps `log` (or `logging`) could be expanded from `:: boolean` to `:: boolean | Keyword.t` to support additional options that would handle these cases.

bruce:
Conspicuously (and accidentally) absent from my previous comment: 

Thanks, @dinsley, for doing this! We'll get in the basic true/false version for you as soon as possible; let's just take a moment to sketch out next steps as well.

dinsley:
No worries!

I have some time I can dedicate to this in order to get it merged, so if there's any changes that you'd like made in order for it to be approved, just let me know.

I think the expansion of `:: boolean` to `:: boolean | Keyword.t` definitely makes sense. The main reason I went with `log` over `logging` for the option was the couple other places that I've disabled the logging in other areas (Phoenix Controllers: https://hexdocs.pm/phoenix/Phoenix.Controller.html) have used `log`.. definitely not opposed to use `logging` though (I actually prefer it, but went with `log` for consistently)

I think even with the scrubbing we would need to disable and use custom logging as we have a pretty particular functionality required for security and reporting purposes.

It might make sense to hook into the Phoenix filter params functionality(`config :phoenix, :filter_parameters, ["password", "api_key"]`) for the scrubbing though as it's already pretty established in Phoenix. Just a thought!

I have some more thoughts around your other points/ideas that I'll follow up on a little bit later today! Just wanted to get these initial thoughts down before I lose them in my head ;)

bruce:
@dinsley Let's go ahead and stick with `:log` for now; consistency, as you say.

Perhaps for filtering we can just provide a reference to a function that would receive arguments/variables and decide whether they should be logged. What say you, @benwilson512 ?

bruce:
I'm going to close this in favor of #61, which will have support for turning off the logging entirely before it's merged. Thank you for bringing this need to our attention and putting in a PR to illustrate it, @dinsley!

PR #60, opened 2017-03-01
update cdn to working fetch polyfill

rphuber:
Looks like the v1.1.0 has been pulled from the cdn.  Updated to the newer version.  all test still pass and fixes issue in browsers not supporting fetch.


bruce:
Thank you; going to cut a quick new version with this in it.

bruce:
Released in v1.2.4.

PR #64, opened 2017-03-14
Fix empty document handling (:no_query_message)

jparise:
Previously, `:no_query_message` wasn't being returned for empty documents.
Instead, we'd get the "No document provider ..." response instead. We
prefer the former in this case because it's more intuitive.

This switches the order of checks within `ensure_processable/2` to inspect
the document before the document provider, restoring the previous "empty
document" response.

### DIFF
--- a/lib/absinthe/plug.ex
+++ b/lib/absinthe/plug.ex
@@ -174,8 +174,8 @@ defmodule Absinthe.Plug do
   @doc false
   @spec ensure_processable(Absinthe.Plug.Request.t, map) :: {:ok, Absinthe.Plug.Request.t} | {:input_error, String.t}
   def ensure_processable(request, config) do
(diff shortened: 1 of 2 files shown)


bruce:
As a document provider can provide a document (that would be set to `nil` otherwise), I don't think this fix is going to be sufficient. I'll pull and experiment a little. I think the fix is going to have to be a bit deeper.

jparise:
That makes sense in concept, but when I looked at the `ensure_processable/2`, it's just calling `ensure_document/2` and `ensure_document_provider/1`, neither of which modify the request.

bruce:
No, I think you're right and this is okay. I had some latent worry that the document provider error would essentially never happen, but I think that in the case where it really matters (a document is given, but there's nothing to process it), we show the right error -- and in other cases, the configured `:no_query_message` is probably good enough.

PR #69, opened 2017-04-04
Batching effort #3

### LINKED ISSUE #34: Add support for transport batching
Now that v1.2 has landed, it would be great to add transport batching support. It was initially contributed in v1.1 by @skosch in #23

benwilson512:
Fixes #34


benwilson512:
@skosch this branch contains the failing test case and the latest batch runner code. Calling the batch runner looks like:
```
prepared_docs |> Batch.Runner.resolve(%{root_value: root_value, schema: schema, context: context})
```

The main challenge from here is integrating the code we came up with previously with the work @bruce did with document providers and stuff. All of the main document preparation and processing stuff we came up with before should be largely reusable, but I just want to make sure it plays nicely with document providers.

skosch:
Neat! Is there anything I can do to help?

benwilson512:
- [ ] Trace the usage of `conn` around. It's used in a lot of places but since it can be altered in an impure way we need to either ensure that it can be returned, or that we pull the bits off we need and pass those around instead.
- [ ] more varied tests.

benwilson512:
- [ ] rework validation result and complexity result phases to return `{:error, bp}` when jumps are not permitted. silently continuing is likely the wrong answer, and makes this very hard.

PR #85, opened 2017-05-29
Update the upload file example.

eamontaaffe:
The current example does not work. The graphql query is not well formed and the name of the query does not match.


benwilson512:
Great catch, thank you!!!

Issue #81, opened 2017-05-09
Batching and upload


yury-dymov:
Hi guys, first of all, thank you for the great library!

I am not sure if it is an issue rather than a question but I would like to be able to upload files and use batching at the same time.

I was playing around with CURL to build the working request first and then to implement a plugin for the apollo-client library to make things work.

For the single request everything works great
```
curl -X POST \
-F query="mutation createPost(\$text: String, \$image: String){createPost(image: \$image, text:\$text, locked: false) { id }}" \
-F operationName="createPost" \
-F variables="{\"text\":\"hello\",\"image\":\"me_jpg\"}" \
-F me_jpg=@me.jpg \
localhost:4000/graphql
```

The question is, how can I convert this single request in a batched one?

Thank you in advance!

benwilson512:
This is now handled on master: https://github.com/absinthe-graphql/absinthe_plug/blob/master/test/lib/absinthe/plug/transport_batching_test.exs#L331-L355

A release will happen soon.

PR #91, opened 2017-06-25
Fixing batching upload issue

yury-dymov:
Please refer https://github.com/absinthe-graphql/absinthe_plug/issues/81

I don't like much the way I fixed the issue but it works fine for me.
I am building my multipart form request with the following body shape:
```
_json: JSON.stringify(Array<Query, Variables>),
file1: File
file2: File
...
```

On the absinthe_plug side, I am checking if `_json` is actually a parsed JSON and if not, parsing it with Poison.

### DIFF
--- a/lib/absinthe/plug/request.ex
+++ b/lib/absinthe/plug/request.ex
@@ -50,18 +50,12 @@ defmodule Absinthe.Plug.Request do
     end
   end
 
-  defp parse_json_params(json) when is_list(json), do: json
-  
-  defp parse_json_params(json), do: Poison.decode!(json)
-
   defp build_request(_body, params, config, batch?: true) do
-    json_params = parse_json_params(params["_json"])
-
-    queries = Enum.map(json_params, fn query ->
+    queries = Enum.map(params["_json"], fn query ->
       Query.parse("", query, config)
     end)
 
-    extra_keys = Enum.map(json_params, fn query ->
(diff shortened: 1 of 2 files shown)


lib/absinthe/plug/request.ex
```diff
@@ -50,12 +50,18 @@ defmodule Absinthe.Plug.Request do
     end
   end
 
+  defp parse_json_params(json) when is_list(json), do: json
+  
+  defp parse_json_params(json), do: Poison.decode!(json)
```
benwilson512: Hey there. This is definitely the right idea, but you can't hardcode any `Poison` calls here. We let people configure their json codec. You can find examples of its use in this module, look for `json_codec`.

benwilson512: This will also need to return a nice error if it does not decode.

benwilson512:
I definitely appreciate you tackling this. There's some nuances involved in order to make it work flexibly, and a test case would be nice as well. If you don't have a chance to handle these bits I'll see if I can take time for it on Monday.

lib/absinthe/plug/request.ex
yury-dymov: Okay, thank you for the insights! 

Will change the implementation accordingly.

yury-dymov:
Implemented changed accordingly, test added.

benwilson512:
Thank you!

PR #92, opened 2017-07-10
Accept functions for the default_{headers,url} graphiql options

djreimer:
The motivation here is to be able to utilize data in `conn` to build up the default headers and default URL for GraphQL.

Use case examples:

- Generate a JSON Web Token for authentication dynamically based on the currently logged in user and append to the default headers.
- Dynamically generate the correct GraphQL endpoint based on current hostname. For example, when subdomains are used for routing, the fully-qualified GraphQL endpoint depends on the current subdomain:

    http://user1.example.com/graphiql -> http://user1.example.com/graphql
    http://user2.example.com/graphiql -> http://user2.example.com/graphql

Thanks for considering!

## Implementation Notes

- I realize there is a bit of duplication in tests now that there are a handful of similar test cases. Happy to do some refactoring, but wanted to propose this with "least invasive" changes to start.
- It seems to me that the function capturing approach is more flexible than a tuple because it allows for easy pattern matching by arity. I didn't spot a clean way to accomplish the same behavior using tuples to denote which function to apply. That being said, I'm relatively new to Elixir so I could be missing something key here!


djreimer:
Hey team, any interest in this?

djreimer:
Hey all, let me know if I should continue resolving merge conflicts on this branch.

benwilson512:
Apologies, I had written a review comment and then forgot to finalize the review!

bruce:
@djreimer If you'd like to give this another peek and address @benwilson512's comment and the merge issue, I think we can get this into the next RC (or the stable release planned for later this week).

djreimer:
Great, will do! Thanks for the bump.

Derrick Reimer
Co-founder, Drip
________________________________

djreimer:
Looks like #108 covered most it! This PR now just adds similar functionality for `default_url`, and increases test and documentation coverage a bit.

benwilson512:
Thank you!

Issue #97, opened 2017-08-10
Batching and file upload using a multipart form


jfrolich:
Loving absinthe so far! 
Just ran in an edge case (I think) when converting my app to support batching. I use a multipart form for submitting the queries and variables when one of the queries contains a file upload. That worked great when I didn't have a batching network interface, but now with batching applied I get the following error: 'no query document found'. I guess it is because normally absinthe expects the body of the document to be an array. However with multipart forms & batching the operations are part of the "operations" field.

PR #98, opened 2017-08-10
fix combination of batching and multipart form

jfrolich:
fixing #97

### DIFF
--- a/lib/absinthe/plug/request.ex
+++ b/lib/absinthe/plug/request.ex
@@ -96,11 +96,21 @@ defmodule Absinthe.Plug.Request do
     end
   end
 
+  defp convert_operations_param(conn = %{params: %{"operations" => operations}}) when is_binary(operations) do
+    put_in(conn.params["_json"], conn.params["operations"])
+    |> Map.delete("operations")
+  end
+  defp convert_operations_param(conn), do: conn
+
   defp extract_body_and_params_batched(conn, body, config) do
-    conn = fetch_query_params(conn)
+    conn =
+      conn
+      |> fetch_query_params()
+      |> convert_operations_param()
+
     with %{"_json" => string} = params when is_binary(string) <- conn.params,
-    {:ok, decoded} <- config.json_codec.module.decode(string) do
-      {:ok, conn, body, %{params | "_json" => decoded}}
+         {:ok, decoded} <- config.json_codec.module.decode(string) do
+         {:ok, conn, body, %{params | "_json" => decoded}}
     else
       {:error, {:invalid, token, pos}} ->
         {:input_error, "Could not parse JSON. Invalid token `#{token}` at position #{pos}"}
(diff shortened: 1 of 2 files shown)


jfrolich:
This is using a slight modification of: https://github.com/jaydenseric/apollo-upload-client (to make it work with absinthe). But any other suggestion for including the operations instead of the "operations" key? I am happy to make it work any other way. Will try to construct a test case!

benwilson512:
Wouldn't it make more sense to just put `operationName` for each item in the `_json` list?

IE if we use this example https://github.com/absinthe-graphql/absinthe_plug/blob/e87efb7688f7afd3618b8492231758c9bd0b4f12/test/lib/absinthe/plug/transport_batching_test.exs#L11-L21

You would have

```json
  [{
    "id": "1",
    "query": "query FooQuery($id: ID!){ item(id: $id) { name } }",
    "variables": {"id": "foo"},
    "operationName":"FooQuery"
  }, {
    "id": "2",
    "query": "query FooQuery($id: ID!){ item(id: $id) { name } }",
    "variables": {"id": "bar"}
  }]
```

jfrolich:
Ah right I just realized the "_json" key does work (I thought before this was a key put there by phoenix that you couldn't send through manually). Is this a better standard? Then no changes are needed. This PR adds the option to also use the "operations" key (added a test as well). This might help compatibility with the rest of the graphql clients otherwise we can close it.

confiks:
@jfrolich ch I'm trying to get Absinthe to work with `apollo-upload-client` myself. I was wondering what "slight modification" you made to the javascript package to get it to work. Could you maybe share the changes?

When trying myself, I noticed that in both `network-interface.js` and `batched-network-interface.js` the requests are put into a multipart-part named 'operations'. In Absinthe (with this pull request), the query and variables seem to be set on the top-level in the non-batching case. Also, Absinthe seems to handle the semantics around naming files in the query and in the multipart-parts differently. Any idea how to reconcile these approaches in Apollo and Absinthe?

confiks:
Scrap that. This pull request plus #101 actually works perfectly fine with `apollo-upload-client`.

To make it work with the `:upload` type out of `Absinthe.Plug.Types`, only a [tiny change](https://gist.github.com/confiks/f627e9a1ed8e9e9342874201d335075a) seems necessary in the `extract-files` dependency of `apollo-upload-client`.

jfrolich:
@confiks: I build a pretty heavy modification also supporting subscriptions... Will look into open-sourcing it!

richeterre:
@confiks, can you elaborate on how you got Absinthe uploads to work with `apollo-upload-client`? Specifically, I'm wondering about two things:

1. Seeing that #101 was dismissed, is it really necessary for this to work?
2. How do you set the mutation's variables? Simply retrieve a `File` or `ReactNativeFile` and set that as a property on the `variables` object?

Would be great to see a short working example of this! I'm using React Native, which is supported by `apollo-upload-client`, so I got really excited when you mentioned that it works 😄

PR #102, opened 2017-08-27
Add subscriptions support to graphiql_workspace

tlvenn:


tlvenn:
This can safely be merged even though the other pieces of the puzzle are not there yet.

I have tested that the query / mutation part is still working as before and the subscription should be working however the absinthe client is running into an error while trying to resolve a promise and I don't understand why. `this.channel` in https://github.com/absinthe-graphql/absinthe-phoenix-js/blob/master/src/index.ts#L53 end up being undefined somehow.

@bruce once graphiql-workspace part is being merged, if you could take a look, it's probably something stupid on my part.

bruce:
@tlvenn Did you want to modify this to use the new binding you have?

bruce:
Update: Talked to @tlvenn; he'll make some edits based on the new JS binding his team has built. If it doesn't make the next rc, we'll get it in and tested before stable.

benwilson512:
@bruce @tlvenn is there a way we could merge this in now and have the updates later? When we do the next book chapter release it would be nice to have subscriptions working over the advanced graphiql UI, since that's the only UI that can make authenticated requests.

tlvenn:
I finally got to the bottom of the issue plaguing the recent version of graphiql-workspace that we need to handle subscriptions.

Once the new version (1.1.3) is released with the fix, this PR can be merged.

tlvenn:
This can be merged now 🎉 
> https://github.com/OlegIlyenko/graphiql-workspace/releases/tag/v1.1.3

PR #104, opened 2017-09-01
Add missing schema field

avitex:
Wondered why the log had schema as nil, turns out it wasn't set anywhere.
For example: `[debug] ABSINTHE schema=nil variables=%{}`


benwilson512:
Thanks!

PR #105, opened 2017-09-01
Fix curl command double quotes escaping

konradkalita:
It is a small doc fix but it is good to have examples that works. ;)

### DIFF
--- a/lib/absinthe/plug/types.ex
+++ b/lib/absinthe/plug/types.ex
@@ -37,7 +37,7 @@ defmodule Absinthe.Plug.Types do
 
   ```shell
   $ curl -X POST \\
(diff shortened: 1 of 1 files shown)


bruce:
@konradkalita Thanks! This actually doesn't work either; this command apparently has been a mess in the docs from the beginning. The `mutation` operation is out of place, and the final curly is missing. Can you replace it with the following? We might as well get rid of the double quotes altogether.

```shell
$ curl -X POST \
  -F query='mutation { uploadFile(users: "users_csv", metadata: "metadata_json") }' \
  -F users_csv=@users.csv \
  -F metadata_json=@metadata.json \
  localhost:4000/graphql
```

benwilson512:
Fixed in master.

PR #106, opened 2017-09-02
Corrected syntax for Plug's forward macro

rawkode:


bruce:
Hmmm.. the following is working just fine in projects we have on Phoenix 1.3 and 1.2.X:

```
forward "/api", Absinthe.Plug,
  schema: OurWeb.Schema,
  # other options
```

The Phoenix docs [also omit](https://hexdocs.pm/phoenix/Phoenix.Router.html#forward/4-examples) a `:to` option.

rawkode:
Hi @bruce, I'm not using Phoenix. Plug requires the `to:`

https://hexdocs.pm/plug/Plug.Router.html#forward/2

bruce:
Right, okay. We're going to have to explicitly lay it out for each, probably; my concern here is most people are using Phoenix and just copy this code out wholesale.

rawkode:
@bruce Yeah, most people will come from Phoenix and won't encounter this issue.

Perhaps just a warning: "If you're not using Phoenix, this syntax will not work. Please consult https://hexdocs.pm/plug/ ?"

yordis:
@bruce @rawkode hah I see why didn't work the first time for me, I was using `Plug` way and I didn't understand why was failing.

@bruce there is some `absinthe_phoenix` wouldn't be better to add the phoenix case there?

benwilson512:
This has been handled in master.

PR #107, opened 2017-09-03
Improve asset system

### LINKED ISSUE #99: GraphiQL Offline Support: GraphiQL Workspace missing fonts
For @yordis: Some leftover work from #88; assets downloaded for GraphiQL Workspace (the default GraphiQL client displayed by Absinthe.Plug) do not include the icon fonts.

avitex:
I had a little play around with the asset system, not sure if I went too far (I originally was just updating the asset versions). My goal was to make it easier to update the assets, unify the assets under one remote source and provide a little more configuration. With `remote_source`, you could swap out, for example to unpkg.

I also have a local commit for moving the config out into mix.

This PR fixes #99

### DIFF
--- a/lib/absinthe/plug/graphiql/assets.ex
+++ b/lib/absinthe/plug/graphiql/assets.ex
@@ -54,27 +54,30 @@ defmodule Absinthe.Plug.GraphiQL.Assets do
       nil ->
         @default_config
       config ->
-        @default_config
-        |> Keyword.merge(Keyword.get(config, :assets, []))
+        Keyword.merge(@default_config, Keyword.get(config, :assets, []))
     end
   end
 
-  def get_assets(), do: reduce_assets(
-    %{},
-    &Map.put(
-      &2,
-      build_asset_path(:local_source, &1),
-      asset_source_url(assets_config()[:source], &1)
+  def get_assets do
+    reduce_assets(
+      %{},
+      &Map.put(
+        &2,
+        build_asset_path(:local_source, &1),
+        asset_source_url(assets_config()[:source], &1)
+      )
     )
-  )
+  end
 
-  def get_remote_asset_mappings, do: reduce_assets(
-    [],
-    &(&2 ++ [{
-      local_asset_path(&1),
-      asset_source_url(:remote, &1)
-    }])
-  )
+  def get_remote_asset_mappings do
+    reduce_assets(
+      [],
+      &(&2 ++ [{
+        local_asset_path(&1),
+        asset_source_url(:remote, &1)
+      }])
+    )
+  end
 
   defp reduce_assets(initial, reducer) do
     Enum.reduce(@assets, initial, fn {package, version, files}, acc ->
@@ -82,8 +85,9 @@ defmodule Absinthe.Plug.GraphiQL.Assets do
     end)
   end
 
-  defp local_asset_path(asset), do: Path.join(
-    assets_config()[:local_directory], build_asset_path(:local_source, asset))
+  defp local_asset_path(asset) do
+    Path.join(assets_config()[:local_directory], build_asset_path(:local_source, asset))
+  end
 
   defp asset_source_url(:smart, asset) do
     if File.exists?(local_asset_path(asset)) do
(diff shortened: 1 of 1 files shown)


yordis:
hah I was checking to fix this one because I forgot 😄

lib/absinthe/plug/graphiql/assets.ex
```diff
@@ -4,54 +4,103 @@ defmodule Absinthe.Plug.GraphiQL.Assets do
...
-      case @config do
-        nil -> []
-        config -> Keyword.get(config, :local_assets, [])
-      end
+    case @config do
+      nil ->
+        @default_config
+      config ->
+        @default_config
+        |> Keyword.merge(Keyword.get(config, :assets, []))
```
yordis: Probably we don't need to use `pipe` here, there is no more than one function call

lib/absinthe/plug/graphiql/assets.ex
```diff
@@ -4,54 +4,103 @@ defmodule Absinthe.Plug.GraphiQL.Assets do
...
+    end
+  end
 
-    Keyword.merge(@default_config, config)
+  def get_assets(), do: reduce_assets(
```
yordis: I would prefer to use `do ... end` block

lib/absinthe/plug/graphiql/assets.ex
```diff
@@ -4,54 +4,103 @@ defmodule Absinthe.Plug.GraphiQL.Assets do
...
+    %{},
+    &Map.put(
+      &2,
+      build_asset_path(:local_source, &1),
+      asset_source_url(assets_config()[:source], &1)
+    )
+  )
+
+  def get_remote_asset_mappings, do: reduce_assets(
```
yordis: I would prefer to use `do ... end` block

lib/absinthe/plug/graphiql/assets.ex
```diff
@@ -4,54 +4,103 @@ defmodule Absinthe.Plug.GraphiQL.Assets do
...
+    Enum.reduce(@assets, initial, fn {package, version, files}, acc ->
+      Enum.reduce(files, acc, &reducer.({package, version, &1}, &2))
+    end)
   end
 
-  def get_assets(:remote), do: @assets
-  def get_assets(:local) do
-    Enum.reduce @assets, %{}, fn asset, acc ->
-      assets_config() |> put_asset_path(asset, acc)
+  defp local_asset_path(asset), do: Path.join(
```
yordis: I would prefer to use `do ... end` block

lib/absinthe/plug/graphiql/assets.ex
```diff
@@ -4,54 +4,103 @@ defmodule Absinthe.Plug.GraphiQL.Assets do
...
+    assets_config()[:local_directory], build_asset_path(:local_source, asset))
+
+  defp asset_source_url(:smart, asset) do
+    if File.exists?(local_asset_path(asset)) do
+      asset_source_url(:local, asset)
+    else
+      asset_source_url(:remote, asset)
     end
   end
+  defp asset_source_url(:local, asset), do: Path.join(assets_config()[:local_url_path], build_asset_path(:local_source, asset))
```
yordis: I would prefer to use `do ... end` block. Also extract some of the values to some variable binding so it's easy to read what is the `Path` joining

lib/absinthe/plug/graphiql/assets.ex
```diff
@@ -4,54 +4,103 @@ defmodule Absinthe.Plug.GraphiQL.Assets do
...
-      |> Path.join(asset_name)
-      |> File.exists?()
-      |> case do
-        true -> assets_config[:url_path] <> "/" <> asset_name
-        false -> asset_external_url
-      end
-
-    Map.put(acc, asset_name, asset_path)
+  defp build_asset_path(source, {package, version, {real_path, aliased_path}}) do
+    assets_config()[source]
```
yordis: I would use some variable binding and then ask for the `[source]`

yordis:
@avitex Niiiiceee! 💟 

I would change some of the code just because it makes it more readable, people will read this code more than actually code it so I think is important, but it looks good to me.

bruce:
@tlvenn If you have any comments on this one, considering your contributions updating the `:advanced` interface, please chime in 😄 

I'm going to look at this in more depth this weekend.

lib/absinthe/plug/graphiql/assets.ex
bruce: I'm not going to comment on every `do:` usage, but our general rule of thumb is that for a given function, all heads should use the same form (either be single line `do:` or `do ... end`).

avitex:
I've made changes as per @yordis and @bruce's suggestions.
Also note I have a local commit for refactoring `asset_config` out into mix.

bruce:
@benwilson512 I'll review this early tomorrow for inclusion in the next rc.

bruce:
Hrmmm,  I can't merge this; looks like the source branch is gone and the patch file is even empty trying to do a manual pull. @avitex any ideas? Looks like your fork disappeared.

PR #108, opened 2017-09-06
Pass connection to default_headers

mgwidmann:
The way I see the default headers function's primary use case is setting an authorization header token for each request. This currently isnt possible without access to the connection.

### DIFF
--- a/lib/absinthe/plug/graphiql.ex
+++ b/lib/absinthe/plug/graphiql.ex
@@ -121,7 +121,9 @@ defmodule Absinthe.Plug.GraphiQL do
         nil -> Map.put(config, :default_headers, "[]")
         {module, fun} when is_atom(fun) ->
           header_string =
-            if(function_exported?(module, fun, 1), do: apply(module, fun, [conn]), else: apply(module, fun, []))
+            module
+            |> function_exported?(fun, 1)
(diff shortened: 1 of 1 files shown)


lib/absinthe/plug/graphiql.ex
```diff
@@ -121,7 +121,7 @@ defmodule Absinthe.Plug.GraphiQL do
         nil -> Map.put(config, :default_headers, "[]")
         {module, fun} when is_atom(fun) ->
           header_string =
-            apply(module, fun, [])
+            if(function_exported?(module, fun, 1), do: apply(module, fun, [conn]), else: apply(module, fun, []))
```
yordis: I would do something like

```
defp call_exported_function(true, module, fun, conn), do: apply(module, fun, [conn])
defp call_exported_function(false, module, fun, _), do: apply(module, fun, [])
```
and then
```
module
|> function_exported?(fun, 1)
|> call_exported_function(module, fun, conn)
...
```

mgwidmann:
Ok, updated as you said. Can you double check the `@spec`, I haven't wrote too many before and want to make sure I did it right...

benwilson512:
Thanks!

PR #110, opened 2017-09-29
Changed documentation to use ~> 1.0.0 plug forward

uri:
I believe the `forward` examples in the documentation uses Phoenix's modified version which has a different API. Since this project has no mention of Phoenix, I believe it makes sense to reference the plug 1.0 API.

### DIFF
--- a/README.md
+++ b/README.md
@@ -69,6 +69,33 @@ forward "/api",
 
 For more information, see the API documentation for `Absinthe.Plug`.
 
+### Phoenix.Router
+
+If you are using [Phoenix.Router](https://hexdocs.pm/phoenix/Phoenix.Router.html), `forward` expects different arguments:
+
+#### Plug.Router
+
+```elixir
+forward "/graphiql",
+  to: Absinthe.Plug.GraphiQL,
+  init_opts: [
+    schema: MyApp.Schema,
+    interface: :simple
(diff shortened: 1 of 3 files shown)


benwilson512:
I do agree that the primary examples need to be aimed at Plug. However, I also think that we should have a `Using in Phoenix` section since the vast majority of people using this are doing so from a phoenix router.

uri:
How about adding a *Pheonix.Router* section to the README and module doc.

Something like:

> If you are using *Phoenix.Router*, `forward` expects different arguments:
```elixir
# Plug.Router
forward "/graphiql",
  to: Absinthe.Plug.GraphiQL,
  init_opts: [
    schema: MyApp.Schema,
    interface: :simple
  ]

# Phoenix.Router
forward "/graphiql",
  Absinthe.Plug.GraphiQL,
   schema: MyApp.Schema,
   interface: :simple
 ```
 
>See https://hexdocs.pm/phoenix/Phoenix.Router.html#forward/4 for details.

uri:
I added a commit with my proposed changes. The updated README can be seen at https://github.com/uri/absinthe_plug/tree/vanilla-plug-opts#phoenixrouter

benwilson512:
Looks good, thank you!

PR #111, opened 2017-10-01
Subscriptions over HTTP

binaryseed:
Hello!

I've put together a demonstration that GraphqQl subscriptions work great over plain HTTP with Server Sent Events (`chunk` responses in `Plug`)..

The core logic is pretty simple, most of this is test code... It'd be quite nice to have subscriptions work in `Absinthe.Plug` out of the box w/o need for WebSockets. These HTTP based subscriptions work from most plain HTTP clients, including good 'ol `curl`..

Let me know what ya'll think! (I know this would need some work before merging...)


lib/absinthe/plug.ex
```diff
@@ -182,6 +186,29 @@ defmodule Absinthe.Plug do
...
+    conn
+    |> put_resp_header("content-type", "text/event-stream")
+    |> send_chunked(200)
+    |> subscribe_loop
+  end
+
+  def subscribe_loop(conn) do
+    receive do
+      %{event: "subscription:data", result: result, topic: _topic} ->
+        case chunk(conn, "#{Poison.encode!(result.data)}\n\n") do
```
benwilson512: We can't hard code the Poison dep here, we need to pass in the config and pull out the json codec from there.

mix.exs
```diff
@@ -40,6 +40,8 @@ defmodule Absinthe.Plug.Mixfile do
       {:plug, "~> 1.3.2 or ~> 1.4"},
       {:absinthe, "~> 1.4.0-rc"},
       {:poison, ">= 0.0.0", only: [:dev, :test]},
+      {:httpoison, "> 0.0.0", only: :test},
```
binaryseed: I'm going to try and figure out a way to test this w/o adding these dependencies...

lib/absinthe/plug.ex
```diff
@@ -182,6 +186,29 @@ defmodule Absinthe.Plug do
     end
   end
 
+  def subscribe(conn, topic, %{context: %{pubsub: pubsub}} = config) do
```
binaryseed: I think that `pubsub` might be worth moving to a top level config item so that it can be set directly in `init`

binaryseed: Maybe re-work a bit of the related logic inside `execute` to make sure it's in the `context` during resolution

binaryseed:
Made that encoding change as well as a handful of others!

benwilson512:
@binaryseed I don't do a lot with chunked responses so this may be totally off base, but is it possible for the client to send down an additional subscription document? That is, if they want to subscribe to several things would they have to send it down all at once (think transport batching) or would they be able to send them down individually?

binaryseed:
So with Server Sent Events, the client only sends the initial request. After that it's one way communication from server -> client.

Does Absinthe support sending multiple `subscription`s in a single query? If so, we could support that, SSE supports "naming" each event that is sent down to the client, which could match the subscription object.

https://developer.mozilla.org/en-US/docs/Web/API/Server-sent_events/Using_server-sent_events

binaryseed:
It seems that it fails if there are 2 subscription fields..

```
query = "subscription {update secondField}"
# ...
(MatchError) no match of right hand side value: [%Absinthe.Blueprint.Document.Field{...
    (absinthe) lib/absinthe/phase/subscription/subscribe_self.ex:73: Absinthe.Phase.Subscription.SubscribeSelf.get_field/1
    (absinthe) lib/absinthe/phase/subscription/subscribe_self.ex:34: Absinthe.Phase.Subscription.SubscribeSelf.get_field_key/2
    (absinthe) lib/absinthe/phase/subscription/subscribe_self.ex:24: Absinthe.Phase.Subscription.SubscribeSelf.do_subscription/1
```

so with SSE as it stands, we just do a single subscription per entire request. this seems fine to me...

bruce:
@binaryseed According to the specification, only one root field is allowed per subscription. https://github.com/facebook/graphql/blob/e80bbfa8640e2cddd7aa6fe30ee1ae64abd7e2e9/spec/Section%205%20--%20Validation.md#single-root-field

We should, however, have a nice validation error instead of what you're encountering.

benwilson512:
Yeah there's a missing validation. If the server can't send additional requests later, we could still probably leverage transport batching and have it send a batch of subscription docs all at once.

binaryseed:
Where would the transport batching be handled? Does the phoenix / websocket variety support this yet?

binaryseed:
Also, the `GraphiQl` plug.. It appears to reference this plug's `call` and `init`, so it should "just work" I suspect. of course the client JS doesn't yet have support for it

benwilson512:
@binaryseed transport batching happens within Absinthe.Plug, it will happen if the batched document structure is used, I'll see if I can draw up a test case and get it to work. Absinthe.Phoenix doesn't need to do this because since it has a persistent connection it can just serially push documents without really any additional overhead.

binaryseed:
Actually, I think that a single subscription per request is a totally fine place to start.. With http/2 on the way, that pretty much means native support for transport batching will come for free

benwilson512:
Well, transport batching isn' simply a matter of using a single HTTP request to send multiple docs, it also lets you perform Absinthe batching across the whole set of documents. That is, if you submit

```
# doc 1
{
  post(id: 1) { comments { body } }
}
# doc 2
{
  post(id: 2) { comments { body } }
}
```

Absinthe will only do a single DB request to get the comments, even though they happen in completely different documents. This requires not only that the requests are sent over a single http connection, but that they are actually sent at the same time.

binaryseed:
That sounds useful for the first event in the subscription, but after that, it's not guaranteed that the will happen at the same time anymore... Comments could be minutes even hours apart

binaryseed:
Ran into a bit of a snag with this... the Browser's `EventSource` API for dealing with Server Sent Events is only able to make `GET` requests :(

This is something the GraphQL spec seems to allow.. http://graphql.org/learn/serving-over-http/#http-methods-headers-and-body

But.. AbsinthePlug seems to dis-allow GET for mutation and subscriptions..

https://github.com/absinthe-graphql/absinthe_plug/blob/master/lib/absinthe/validation/http_method.ex

Is there a specific reason for that? Seems like subscriptions should be fine over `GET`

binaryseed:
Tweaked it to do just that!

binaryseed:
Are ya'll interested in getting this in for 1.4? Are there code changes you'd like to see? specs, moving stuff, etc?

benwilson512:
Hey @binaryseed sorry, hadn't had a chance to look at it after the last commit.

benwilson512:
This looks great, thank you!

PR #119, opened 2017-11-06
Chunk bug fix

benwilson512:


lib/absinthe/plug.ex
```diff
@@ -224,24 +224,26 @@ defmodule Absinthe.Plug do
...
         case chunk(conn, "#{encode_json!(result, config)}\n\n") do
           {:ok, conn} ->
-            subscribe_loop(conn, config)
+            subscribe_loop(conn, topic, config)
           {:error, :closed} ->
+            Absinthe.Subscription.unsubscribe(config.context.pubsub, topic)
             conn
         end
       :close ->
+        Absinthe.Subscription.unsubscribe(config.context.pubsub, topic)
```
benwilson512: @binaryseed Hey! So I'm not finding situations where I ever get a `:close` message, at least when using CURL. Possibly I'm just doing CURL wrong, but I'd like to be able to hook in an unsubscribe call in here somewhere. It isn't a huge deal but Cowboy doesn't always kill the processes, and they can sort of hang around for a while. In such cases their memory just grows while we push events at them, which seems no good.

binaryseed: Yeah, I put that in there to enable the tests... I did a bunch of digging into server sent events, and given that they are 1 way communication, you can only detect that the connection is closed when you try to `chunk` and the response comes back as `{:error, :closed}`. 

I couldn't really trigger that condition in the tests because plug tests aren't really a full http connection, hence this way of manually triggering the receive loop to close.

I've also been thinking about this case. Makes sense to put the `unsubscribe` in. I was considering a configurable `after` timeout that would send a "comment" chunk, essentially a keep-alive ping. That would give the server a chance to detect a dropped connection even if it's not trying to send any updates...

This strategy is mentioned here: https://developer.mozilla.org/en-US/docs/Web/API/Server-sent_events/Using_server-sent_events

benwilson512: Awesome. I'll address that in a separate PR then.

PR #124, opened 2017-11-21
Make downloadable playground fonts

avitex:
Fix for #99

- Make playground fonts downloadable
- Clean up templates


avitex:
No problems @tlvenn. Thanks for adding playground, it's a very nice interface!

lib/absinthe/plug/graphiql/graphiql_workspace.html.eex
```diff
@@ -44,7 +41,7 @@ add "&raw" to the end of the URL within a browser.
 
     ReactDOM.render(
       React.createElement(graphiqlWorkspace.GraphiQLWorkspace, {config: config}),
-      document.getElementById('workspace')
+      document.body
```
yordis: @tlvenn @avitex  are you sure this was a good idea? There is some other elements in the `body` like `script` tag so I am not sure if this is good for React ...

avitex: Yeah, it's fine. Browsers store code in memory (do not unload), and we are not using the returned reference from `ReactDOM.render`

yordis: > Don’t Render to Body
So what’s the other mistake I wanted to tell you about? It’s much simpler: don’t ever render to document.body. A lot of React examples do that because it’s less typing and looks more clear. Don’t. Do. This.

From Dan Abramov, article https://medium.com/@dan_abramov/two-weird-tricks-that-fix-react-7cf9bbdef375

I have browser extensions that inject elements to the body right now, like `ghostery`

tlvenn: Ya good point @yordis, seems like we really should not mess with the body. @avitex do you mind creating a small PR to fix this ?

avitex: Sure. GraphiQL will need to be changed too then, as it was previously rendering to `document.body` too

PR #132, opened 2017-12-06
ensure the socket URL is a string

craig-day:
When I tried to add a custom websocket URL, I ended up seeing this error in my javascript console: 

```
Uncaught SyntaxError: Unexpected token :
```

when I click through the error, it looks like the websocket URL isn't a string. 

screen shot 2017-12-06 at 3 23 50 pm

I actually found this out when working on #131, but in the interest of simple isolated fixes I pulled it into a separate PR. If you decide to merge #131 right away then this fix doesn't really matter. 

/cc @bruce @benwilson512 @tlvenn


tlvenn:
Good catch @craig-day , since #131 has been merged, I will close this PR.

PR #136, opened 2017-12-21
Enquote socket url to prevent it from breaking js

### LINKED ISSUE #135: Using socket_url for workspace causes a js syntax error
Using graphiql workspace with the `socket_url` option will prevent graphiql workspace from rendering in the browser because of a js syntax error.

DavidJFelix:
-`socket_url` is a string, so it will be templated literally into the template string
- We want it in quotes, so we have to put quotes around it.
- Fixes #135

### DIFF
--- a/lib/absinthe/plug/graphiql/graphiql_workspace.html.eex
+++ b/lib/absinthe/plug/graphiql/graphiql_workspace.html.eex
@@ -32,7 +32,7 @@ add "&raw" to the end of the URL within a browser.
     var config = new graphiqlWorkspace.AppConfig("graphiql", {
       defaultUrl: <%= default_url %>,
       <%= if socket_url do %>
-        defaultWebsocketUrl: <%= socket_url %>,
+        defaultWebsocketUrl: '<%= socket_url %>',
         subscriptionsClientBuilder: absintheSubscriptionsClientBuilder,
       <% end %>
       defaultQuery: '<%= query_string %>',


benwilson512:
The problem is, we do want it to be a JS literal: https://github.com/absinthe-graphql/absinthe_plug/blob/master/lib/absinthe/plug/graphiql.ex#L372

DavidJFelix:
Using it in phoenix right now and this is what happens:

```
defmodule ApiWeb.Router do
  use ApiWeb, :router

  pipeline :api do
    plug :accepts, ["json"]
  end

  scope "/" do
    pipe_through :api
    
    forward "/graphiql",
      Absinthe.Plug.GraphiQL,
        schema: ApiWeb.Schema,
        default_url: "http://localhost:4000/graph",
        socket_url: "ws://localhost:4000/socket"

    forward "/graph", Absinthe.Plug, schema: ApiWeb.Schema
  end

end
```

```
<script type="text/javascript">
    function absintheSubscriptionsClientBuilder(url, connectionParams) {
      return new AbsintheSocketGraphiql.SubscriptionsClient(url, {params: connectionParams});
    }

    var protocol = window.location.protocol === "https:" ? "wss:" : "ws:";
    var config = new graphiqlWorkspace.AppConfig("graphiql", {
      defaultUrl: 'http://localhost:4000/graph',
      
        defaultWebsocketUrl: ws://localhost:4000/socket,
        subscriptionsClientBuilder: absintheSubscriptionsClientBuilder,
      
      defaultQuery: '',
      defaultVariables: '',
      defaultHeaders: []
    });

    ReactDOM.render(
      React.createElement(graphiqlWorkspace.GraphiQLWorkspace, {config: config}),
      document.getElementById("root")
    );
  </script>
```

It doesn't wrap defaultWebsocketUrl in quotes.

DavidJFelix:
However... I don't think these even work. I modified the socket_url to be `"'ws://localhost:4000/socket'"` and it populates that value into the page correctly and renders, but the field is not populated in the UI.

Custom default URL is also ignored.

DavidJFelix:
Okay. So I did a little more searching and they do populate the page correctly if I clear localstorage before hand. My current workaround is:

```elixir
defmodule ApiWeb.Router do
  use ApiWeb, :router

  pipeline :api do
    plug :accepts, ["json"]
  end

  scope "/" do
    pipe_through :api
    
    forward "/graphiql",
      Absinthe.Plug.GraphiQL,
        schema: ApiWeb.Schema,
        default_url: "http://localhost:4000/graph",
        socket_url: "'ws://localhost:4000/socket'"

    forward "/graph", Absinthe.Plug, schema: ApiWeb.Schema
  end

end
```
In my router (notice the double and single quotes on `socket_url`). If the url changes, I have to clear localstorage. This may be something to fix in graphiql-workspace, but the quote seems to be related to the way that this plug is rendering the eex file.

PR #143, opened 2018-02-08
makes socket_url configurable via function as docs suggest

fredr:
According to the docs, `:socket_url` can be a function: https://github.com/absinthe-graphql/absinthe_plug/blob/master/lib/absinthe/plug/graphiql.ex#L114-L125

This probably also solves #135, because of: https://github.com/absinthe-graphql/absinthe_plug/compare/master...fredr:fix/graphiql_socket_url?expand=1#diff-c40bc49b61fa68956d5b9309edae59a4R379

This way both `:socket_url` and `:default_url` works the same way.

### DIFF
--- a/lib/absinthe/plug/graphiql.ex
+++ b/lib/absinthe/plug/graphiql.ex
@@ -191,34 +191,48 @@ defmodule Absinthe.Plug.GraphiQL do
     end
   end
 
+  defp function_arity(module, fun) do
+    Enum.find([1, 0], nil, &function_exported?(module, fun, &1))
+  end
+
+  defp get_config_val(config, key, conn) do
+    case Map.get(config, key) do
+      {module, fun} when is_atom(fun) ->
+        case function_arity(module, fun) do
+          1 -> apply(module, fun, [conn])
+          0 -> apply(module, fun, [])
+          _ ->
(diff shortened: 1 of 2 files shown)


lib/absinthe/plug/graphiql.ex
```diff
@@ -221,6 +221,20 @@ defmodule Absinthe.Plug.GraphiQL do
           raise "invalid default url: #{inspect val}"
        end
 
+     config = case config[:socket_url] do
+        nil -> config
+        val when is_binary(val) -> Map.put(config, :socket_url, val)
+        {module, fun} when is_atom(fun) ->
+          url =
+            module
+            |> function_exported?(fun, 1)
```
benwilson512: I would prefer if we raised here instead of just ignoring it in the `call_exported_function` function.

fredr: sure! Do you want me to do that in `call_exported_function`? so that its the same for `default_headers`, `default_url` and `socket_url`?

fredr: @benwilson512 I've updated! its a bit messy, since both 1 or 0 arity is supported, and for the socket stuff, there is also the `socket` config option

benwilson512:
Thanks!

PR #149, opened 2018-03-18
add required init function for Plug.Parser behaviour

mmmries:
I noticed that after an update to plug `1.5.0` I was getting a compile-time warning like this

```
==> absinthe_plug
Compiling 18 files (.ex)
warning: function init/1 required by behaviour Plug.Parsers is not implemented (in module Absinthe.Plug.Parser)
  lib/absinthe/plug/parser.ex:1
```

It looks like this callback used to be optional, but is now required. I tested compiling under the current `mix.lock` and after a `mix deps update --all` to make sure it compiles cleanly under both.

### DIFF
--- a/lib/absinthe/plug/parser.ex
+++ b/lib/absinthe/plug/parser.ex
@@ -21,6 +21,9 @@ defmodule Absinthe.Plug.Parser do
   alias Plug.Conn
 
   @doc false
+  def init(opts), do: opts
+
+  @doc false
   def parse(conn, "application", "graphql", _headers, opts) do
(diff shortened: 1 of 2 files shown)


mmmries:
I'm looking to contribute to absinthe a bit more overall, would you as maintainers be interested in PRs to
* test a wider variety of elixir/OTP versions on Travis?
* use the standard elixir formatter and check formatting during CI?
* add more doctests or docs?

If so I would be happy to open some PRs toward that end.

mmmries:
ping

PR #150, opened 2018-04-02
Added checking conn state before sending resp.

aliksend:
This change is small, but it is very useful when you want to `send_resp` in before_send hook.

About my situation. 
I want to check authorization just before resolve function, because my app uses users permissions to allow/deny access to fields. So I use 
```elixir
    Absinthe.Plug.put_options(conn, context: %{ current_user: resource })
```
in plug before `Absinthe.Plug` and then (in `before_send`) I want to check that authorization is checked (I can forget to add auth checker middleware before resolve func) and passed (is user authorized and have required permissions) and **send 401 response with message** if authorization failed. But if I try to do it, I get `Plug.Conn.AlreadySentError` in `(absinthe_plug) lib/absinthe/plug.ex:471: Absinthe.Plug.encode/4` because absinthe_plug tries to send_resp second time.
Also I think that it can be useful in other situations.
Thanks

### DIFF
--- a/lib/absinthe/plug.ex
+++ b/lib/absinthe/plug.ex
@@ -239,35 +239,39 @@ defmodule Absinthe.Plug do
     config = update_config(conn, config)
     {conn, result} = conn |> execute(config)
 
-    case result do
-      {:input_error, msg} ->
-        conn
-        |> send_resp(400, msg)
+    if conn.state == :unset do
+      case result do
+        {:input_error, msg} ->
+          conn
+          |> send_resp(400, msg)
 
-      {:ok, %{"subscribed" => topic}} ->
-        conn
-        |> subscribe(topic, config)
+        {:ok, %{"subscribed" => topic}} ->
+          conn
+          |> subscribe(topic, config)
 
-      {:ok, %{data: _} = result} ->
-        conn
-        |> encode(200, result, config)
+        {:ok, %{data: _} = result} ->
+          conn
+          |> encode(200, result, config)
 
-      {:ok, %{errors: _} = result} ->
-        conn
-        |> encode(400, result, config)
+        {:ok, %{errors: _} = result} ->
+          conn
+          |> encode(400, result, config)
 
-      {:ok, result} when is_list(result) ->
-        conn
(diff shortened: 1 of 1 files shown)


benwilson512:
Hey, thanks for the PR!

Unfortunately I'm not entirely sure that I'm on board with the use case. If you're running the GraphQL document, you should return a GraphQL response which would be whatever Absinthe.Plug wants to return. If you aren't running the document you can just halt from an earlier plug.

From a technical perspective, this PR could definitely use a test case to illustrate the desired flow. I'm actually a bit skeptical that this PR fixes the issue that you think it does. `before_send` callbacks are gonna be run exactly when  `Absinthe.Plug` calls `send_resp`. This is to say, the `conn.state` passed into the Absinthe.Plug will always be `unset` because none of the `before_send` callbacks have run yet. 

> when you want to send_resp in before_send hook

I'm also skeptical that this is a good idea. When `before_send` callbacks run it's because some other plug has _already_ called `send_resp`, and now the callback is trying to hijack that.

aliksend:
Thank you for your response.

I agree with you that returning graphql response is better idea, for example for situations when user have access to one fields and haven't to others, and he can get `data` for that one fields and `errors` with `path` for another. And successfully processed graphql request should return 200 even if it contains errors. This is more correct in terms of graphql but more difficult to implement both on backend and on frontend.
May be I really should find another solution. I played some with absinthe middleware callbacks, but couldn't get wanted behavior.
So I'll will try to implement it in more suitable way and reply to this PR if find another advantages of proposed solution.

About control flow: I found that `before_send` callbacks runs before `send_resp`.
Call stack looks like `call/2` –> `execute/1` –> `run_request/4` –> `run_query/4` –> `apply_before_send/3` –> `apply/3`. And `send_resp` is called in `call/2` method after calling `execute` so that also after `before_send` callback.

benwilson512:
Ah, you're using the `before_send` option of Absinthe.Plug, not the `Plug.Conn.register_before_send`. Yes the flow is correct then. Nonetheless, I don't think it's a right use of that feature to try to override the response that Absinthe is going to send.

PR #155, opened 2018-05-18
Fix compatibility with Phoenix master

rafaelgaspar:
On Phoenix master branch `Endpoint.__sockets__` is now returning a 4 elements tuple instead of 2, so I changed the anonymous function the filters it out to work with any size tuple.


benwilson512:
Thanks!

PR #156, opened 2018-06-06
On errors return 200 so it plays nice with Apollo

hubertlepicki:
This relates to: https://github.com/absinthe-graphql/absinthe/issues/554

In nutshell: Apollo.js assumes that we get 200 OK even on errors. GraphQL specification does not seem to care what status codes are in use (at least I can't find it), so I *think* it might be reasonable to assume it's always 200 OK and errors are handled on the layer of protocol rather than transport (HTTP) unless catastrophic failure happens (500 & friends).

This pull request changes the current behavior of returning 400 in case "errors" are set, to returning 200.

### DIFF
--- a/lib/absinthe/plug.ex
+++ b/lib/absinthe/plug.ex
@@ -254,7 +254,7 @@ defmodule Absinthe.Plug do
 
       {:ok, %{errors: _} = result} ->
         conn
-        |> encode(400, result, config)
+        |> encode(200, result, config)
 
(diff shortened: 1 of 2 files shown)


yordis:
@hubertlepicki maybe we should add a config to the plug reading the desired status code (using `400` as default since some folks could have some breaking changes)

mydearxym:
ping ..

PR #160, opened 2018-07-02
WIP: Update to playground 1.6.1

tlvenn:
Now that Playground with the 1.6.X release is supporting custom apollo link, there is no reason anymore to fork it. Unfortunately, we can't use the html version they are providing as it does not support passing a custom apollo link...

So I created a mini react app that bundles GraphQL Playground with a custom link compatible with Apollo.

Everything should work but I have yet to test subscriptions.


tlvenn:
@benwilson512 @bruce if you get the chance to give it a try, that would be great.

PR #161, opened 2018-07-11
pass query-string params to websocket

tim2CF:
Before:
screen shot 2018-07-11 at 16 45 38
Produced error
```
'ws://127.0.0.1:4000/graphql/backoffice/chat?s=1&sf=chat/websocket&vsn=2.0.0' failed: Error during WebSocket handshake: Unexpected response code: 404
```
Because URL was passed to phoenix.js as it was, resulted URL was incorrect (problem with `/websocket` concat)

This PR resolve this issue and query string is passed as `params` option field


benwilson512:
Hey this looks great! Does a similar thing need to happen with the graphiql playground ui?

tim2CF:
Hi! Thanks)
I don't know about graphiql playground, I'm pretty new with graphql, just solved the problem that faced today

knewter:
I can confirm the same thing happens on graphql playground cc @tlvenn, I just ran into it

PR #163, opened 2018-07-27
Update README.md

Adzz:
This confused me (easily done) maybe it will help others?


bruce:
I think more clearly calling out the router file is a good idea, but the `Plug.Parsers` code above is _also_ in that file (above that portion, we say "configure your router like:"

Maybe change that instruction instead to be clearer?

Adzz:
Sorry I'm not understanding, do you put the plug code inside your `router.ex` file rather than in `endpoint.ex`?

PR #165, opened 2018-07-28
Add source url to mix.exs

maartenvanvliet:
Add source url to mix.exs, this way the generated documentation can link to the source code on github.


benwilson512:
Thanks!

PR #178, opened 2018-09-11
Reject non-batch JSON payloads

emeryotopalik:
It is possible to send a non-batch query as a json string. For example:
```
"{
    \"query\": \"{ query }\"
}"
```
This causes a BadMapError to fire at https://github.com/absinthe-graphql/absinthe_plug/blob/master/lib/absinthe/plug/request/query.ex#L132 because it is getting a tuple from the `Enum.map` here: https://github.com/absinthe-graphql/absinthe_plug/blob/master/lib/absinthe/plug/request.ex#L55

The solution is to treat a non-batch json payload as a regular query once it is properly extracted.

### DIFF
--- a/lib/absinthe/plug/request.ex
+++ b/lib/absinthe/plug/request.ex
@@ -45,7 +45,7 @@ defmodule Absinthe.Plug.Request do
       # Phoenix puts parsed params under the "_json" key when the
       # structure is an array; otherwise it's just the keys themselves,
       # and they may sit in the body or in the params
-      batch? = Map.has_key?(params, "_json") && is_list(params["_json"])
+      batch? = Map.has_key?(params, "_json")
       {:ok, conn, build_request(body, params, config, batch?: batch?)}
     end
   end
@@ -65,7 +65,6 @@ defmodule Absinthe.Plug.Request do
       extra_keys: extra_keys,
     }
   end
-
   defp build_request(body, params, config, batch?: false) do
     queries =
       body
--- a/lib/absinthe/plug/request.ex
+++ b/lib/absinthe/plug/request.ex
@@ -45,7 +45,7 @@ defmodule Absinthe.Plug.Request do
       # Phoenix puts parsed params under the "_json" key when the
       # structure is an array; otherwise it's just the keys themselves,
       # and they may sit in the body or in the params
-      batch? = Map.has_key?(params, "_json")
+      batch? = Map.has_key?(params, "_json") && is_list(params["_json"])
       {:ok, conn, build_request(body, params, config, batch?: batch?)}
     end
   end
@@ -65,6 +65,19 @@ defmodule Absinthe.Plug.Request do
       extra_keys: extra_keys,
     }
   end
+
+  defp build_request(body, %{"_json" => query}, config, batch?: false) do
+    queries =
+      body
+      |> Query.parse(query, config)
+      |> List.wrap
+
+    %__MODULE__{
+      queries: queries,
+      batch: false,
+    }
+  end
+
   defp build_request(body, params, config, batch?: false) do
     queries =
       body
--- a/lib/absinthe/plug/request.ex
+++ b/lib/absinthe/plug/request.ex
@@ -66,18 +66,6 @@ defmodule Absinthe.Plug.Request do
     }
   end
 
-  defp build_request(body, %{"_json" => query}, config, batch?: false) do
-    queries =
-      body
-      |> Query.parse(query, config)
-      |> List.wrap
-
-    %__MODULE__{
-      queries: queries,
-      batch: false,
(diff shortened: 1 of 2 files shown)


benwilson512:
Non batched JSON is fully supported, it just expects a top level `query` key. You're using a top level `_json` key. It seems like your client is misconfigured if it isn't sending the query key at the top level

test/lib/absinthe/plug_test.exs
```diff
@@ -107,6 +107,18 @@ defmodule Absinthe.PlugTest do
     assert resp_body == @foo_result
   end
 
+  test "content-type application/json _json query works" do
+    opts = Absinthe.Plug.init(schema: TestSchema)
+    json_query = Poison.encode!(%{"_json" => %{"query" => @query}})
```
benwilson512: Why is `_json` here?

emeryotopalik:
After further digging, I think this might actually be a bug in Plug. "_json" is only supposed to be attached to arrays https://github.com/elixir-plug/plug/blob/master/lib/plug/parsers/json.ex#L5, but it's not handling strings properly here: https://github.com/elixir-plug/plug/blob/master/lib/plug/parsers/json.ex#L55-L62

I'm going to close this and file a ticket with Plug.

emeryotopalik:
@benwilson512 reopening this. Turns out Plug used some misleading words in their documentation about parsing JSON. https://github.com/elixir-plug/plug/blob/master/lib/plug/parsers/json.ex#L5 Non-array values can be parsed into the `_json` object. Those need to be properly handled here. I updated the test case to a more realistic query string that gets encoded and parsed into `_json` by plug. 

The alternative solution would be to always reject string queries.

test/lib/absinthe/plug_test.exs
```diff
@@ -107,6 +107,20 @@ defmodule Absinthe.PlugTest do
     assert resp_body == @foo_result
   end
 
+  test "content-type application/json _json query works" do
+    opts = Absinthe.Plug.init(schema: TestSchema)
+    query = """
+    { "query": "{ item(id: \\"foo\\") { name } }" }
+    """
+
+    assert %{status: 200, resp_body: resp_body} = conn(:post, "/", Poison.encode!(query))
```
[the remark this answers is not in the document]
emeryotopalik: Yah, I'll provide the example that we were able to get through on. But it was essentially this.

benwilson512:
I guess I don't understand the purpose of this PR. If someone sends a double encoded JSON string to Absinthe, it should not handle it, I don't know why it would.

lib/absinthe/plug/request.ex
```diff
@@ -65,6 +65,19 @@ defmodule Absinthe.Plug.Request do
       extra_keys: extra_keys,
     }
   end
+
+  defp build_request(body, %{"_json" => query}, config, batch?: false) do
+    queries =
+      body
+      |> Query.parse(query, config)
```
benwilson512: I don't understand why Absinthe should do this. What if it's triple encoded, quadruple? The client should just send the correct request body.

emeryotopalik: Well, at the point at which you get to triple encoding, your decoder will raise an error. I just want to deal with the fact that you get a `BadMapError` because you attempt to iterate over string in the case that someone sends a valid JSON string.

emeryotopalik:
As I mentioned - I also think completely rejecting JSON strings is a valid solution.

benwilson512:
Right, I agree that the current error message isn't great. However I do think that an error message is the correct response, since a double encoded string isn't a valid GraphQL HTTP request.

emeryotopalik:
@benwilson512 does this seem better to you?

test/lib/absinthe/plug_test.exs
```diff
@@ -107,6 +107,20 @@ defmodule Absinthe.PlugTest do
     assert resp_body == @foo_result
   end
 
+  test "content-type application/json fails with improperly encoded query strings" do
+    opts = Absinthe.Plug.init(schema: TestSchema)
+    query = """
+    { "query": "{ item(id: \\"foo\\") { name } }" }
```
binaryseed: What is the simplest post body that can generate this condition?

would something like this do it?

```json
"foo"
```

emeryotopalik: It would need the query object, or it gets caught earlier on. So something like: 
`"{\"query\": \"\"}"`

binaryseed:
I spent some time on this issue today and I think we have a nice summary and solution over in https://github.com/absinthe-graphql/absinthe_plug/pull/186

benwilson512:
Handled via https://github.com/absinthe-graphql/absinthe_plug/pull/186

Issue #169, opened 2018-08-08
Batch query mismatch between apollo and absinthe


qqdanielqq:
Following on from our conversation on the slack channel, we are encountering a mis match between apollo making batch queries and the response that absinthe gives us.

Response from Apollo Server
```[{"data":{"__schema":{"queryType":{"name":"RootQuery"}}}},{"data":{"__schema":{"queryType":{"description":""}}}}]```

Response from Absinthe
```[{"operationName":"_fooexample","payload":{"extensions": ...,"data": ... }}]```

Attached is a small demo project that demonstrates this:

https://github.com/kirill-kruchkov/apollo-client-batch-n-state-issue-demo

If you can provide any help with this that would be great!!

**Comment by thomasbrus** [comment]
If you're using Apollo Client in combination with `BatchHttpLink` you can simply workaround the issue as follows:

```js
const absintheAfterwareLink = new ApolloLink((operation, forward) => {
  return forward(operation).map(response => {
    // Also see https://github.com/absinthe-graphql/absinthe_plug/issues/89#issuecomment-307845018
    return { ...response, data: response.payload.data };
  });
});

let link = ApolloLink.from([
  absintheAfterwareLink,

  // ...,

  new BatchHttpLink({
    batchMax: 100,
    batchInterval: 10,
  }),
]);

let cache = new InMemoryCache();

export default new ApolloClient({ link, cache });
```

binaryseed:
Addressed in https://github.com/absinthe-graphql/absinthe_plug/pull/237

PR #179, opened 2018-09-18
Change response shape to work together with apollo-link-batch-http

aleqsio:
It seems that the new version of apollo client expects the data to not be wrapped in ```payload```, and current format causes errors with writing to cache.

### DIFF
--- a/lib/absinthe/plug.ex
+++ b/lib/absinthe/plug.ex
@@ -404,9 +404,7 @@ defmodule Absinthe.Plug do
       results
       |> Enum.zip(request.extra_keys)
       |> Enum.map(fn {result, extra_keys} ->
-        Map.merge(extra_keys, %{
-          payload: result
-        })
+        Map.merge(extra_keys, result)
       end)
 
     {conn, {:ok, results}}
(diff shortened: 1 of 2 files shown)


binaryseed:
cc @mbaker

aleqsio:
Hi,
sorry for the long delay (almost a year 😄).
I've added a `use_batch_http_link_format` boolean to the config, a test to see if this works correctly, ensured the CI passes, ect.
This solves #169.
I can also update the docs for this one, just wanted to wait for your approval.

binaryseed:
Hello!

Ideally this would have a documentation note...

Also, I'm wondering about the name - given that it's a config option whatever we choose will stick around forever.

Do you think we could name it something that is more directly understandable? `batch_payload_key`?

FRSgit:
Are there any updates on this topic? I think it's an important thing to have a server compatible with Apollo response structure.

@benwilson512 @binaryseed What is left to be done here to get this PR merged? Only the documentation info?

@binaryseed I like the new name 👍

binaryseed:
Merging this into a working branch to make a few tweaks

PR #180, opened 2018-09-21
Point to the correct json config option

binaryseed:
I noticed that there are two config keys floating around for the JSON setting, one of which seems to be incorrect...

I changed all instances of `json_decoder` to `json_codec`. This probably worked for most folks since it defaults to `Poison`


benwilson512:
Ouch, don't know how that happened, thanks!

binaryseed:
hmm, I might have messed something up here

PR #181, opened 2018-09-21
Revert "Point to the correct json config option"

binaryseed:
I think this PR was wrong, it confused the `plug` config option with `the `absinthe_plug` config option. Sorry about that!!

Reverts absinthe-graphql/absinthe_plug#180


benwilson512:
Hey yeah I jumped the gun on the merge, made the same mistake you did. The reason it's called `json_decoder` for `Absinthe.Plug` is that `Absinthe.Plug` only _decodes_ JSON, it never needs to encode it itself.

PR #184, opened 2018-10-13
Fix finding WebSocket path from Phoenix Socket

jayjun:
In Phoenix 1.4, `MyApp.Endpoint.__sockets__` returns `{path, module, opts}` tuples instead of `{path, module}`.

`Absinthe.Plug.GraphiQL.find_socket_path/2` should be able to handle both .


benwilson512:
Thanks!

PR #190, opened 2018-11-14
Always return a JSON body

### LINKED ISSUE #176: Default `no_query_message` to JSON that meets GraphQL spec
Howdy.. currently the default error response body for an empty query document is a plain string. This isn't friendly for clients that are expecting JSON.

https://github.com/absinthe-graphql/absinthe_plug/blob/master/lib/absinthe/plug.ex#L172

I'm suggesting that we change it to return `application/json` & a JSON payload that meets the GraphQL spec

```json
{
  "errors": [{"message": "Foo"}]
}
```

The fix would be rendering any `{:input_error, msg}` through `encode`

https://github.com/absinthe-graphql/absinthe_plug/blob/master/lib/absinthe/plug.ex#L247-L249

I'll submit a PR if ya'll think this should change...

binaryseed:
This PR ensures that we always return a JSON response for any request. 

The GraphQL spec states that the response should always be a map: https://facebook.github.io/graphql/June2018/#sec-Response-Format

Also, clients will often try to decode the response body as JSON and when it's raw text that will fail.

This would technically change existing behavior, so I'm open to feedback...

@benwilson512 
Fixes #176

### DIFF
--- a/lib/absinthe/plug.ex
+++ b/lib/absinthe/plug.ex
@@ -249,7 +249,7 @@ defmodule Absinthe.Plug do
     case result do
       {:input_error, msg} ->
         conn
-        |> send_resp(400, msg)
+        |> encode(400, error_result(msg), config)
 
       {:ok, %{"subscribed" => topic}} ->
         conn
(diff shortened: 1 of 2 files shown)


benwilson512:
Great, thanks!

PR #191, opened 2018-11-18
Make poison dependency optional

BleuShan:
There's no point to require it on the clients since it's configurable.


benwilson512:
If you go to https://hex.pm/packages/absinthe_plug you'll see that `Poison` isn't actually required. Hex only counts prod dependencies. Poison is just required in dev and test to run the test suite.

PR #197, opened 2019-02-12
allow to set default query

ayrat555:
Changelog:
- allow passing `default_query` parameter to `advanced` GraphiQL interface

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@ v1.5.0
 
 - Basically no changes in Absinthe.Plug, but required for Absinthe 1.5 pre-release support
 - Chore: Update Plug. Get rid of plug compilation warnings
+- Feature: allow to pass a default query to GraphiQL interface
 
 v1.4.6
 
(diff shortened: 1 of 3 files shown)


ayrat555:
@benwilson512 @bruce can you take a look when you have time? thanks

ayrat555:
@benwilson512 ping

ayrat555:
@bruce you're my only hope!

axelson:
@ayrat555 just curious, what is the point of passing a `default_query`? i.e. what does it do? What is a typical use-case. Also I would think that it would be best to support the basic graphiql as well as advanced.

ayrat555:
@axelson to show it as an example in the graphiql interface. `playground` does not support passing default query. There is an issue https://github.com/prisma/graphql-playground/issues/301

benwilson512:
Hi @ayrat555, please use `mix format` so that the CI passes.

ayrat555:
@benwilson512 done

ayrat555:
@benwilson512 will you merge it?

ayrat555:
thanks!

PR #201, opened 2019-04-10
[WIP] Enable to upload multiple files as list_of field

### LINKED ISSUE #139: Multiple files upload
Hey guys!
Does the library support multiple file uploading? If so could you please provide with an example of schema and especially an example of a request. Thanks.

inaniyants:
### Background:
When application allows to upload big amount of files, there is few options how this could be implemented: 
* option1:  define mutation with multiple upload arguments (the number of arguments could be 10+)
* option2: schedule an upload queue on a client and execute upload mutation for each file separately
* option3: use batching and run multiple upload mutations 

In my opinion all 3 options are not comfortable to develop and support.

Alternative solution could be next mutation: 
```
  object :file_mutations do
    field :upload_files, list_of(:file) do
      arg :files, non_null(list_of(non_null(:upload)))

      resolve(&Resolvers.File.handle_files_upload/2)
    end
  end
```

### Problem:
Such alternative approach currently is not supported by absinthe.

### Suggested in this PR solution
Imagine a new convention: when list_of uploads is used for `:files` argument, then client should prefix an argument value with `[]` (prefix is chosen in order to pattern match string) and name each file in form data with `[n]` suffix (where n is a file index in list).
```
$ curl -X POST \\
-F query="{mutation uploadFile(users: \"[]files\", metadata: \"metadata_json\")" \\
-F files[0]=@users.csv \\
-D files[1]=@currencies.csv \\
-F metadata_json=@metadata.json \\
localhost:4000/graphql
```

Closes #139 .

### WIP 
My current solution has a problem: since `:upload type` return a list and argument `:files` is a list_of, resolver receives list of `Plug.Upload` wrapped into one more list. See test.
```
%{
  files: [
    [
      %Plug.Upload{
        content_type: "application/octet-stream",
        filename: "users.csv",
        path: "/tmp/random_name.csv"
      },
      %Plug.Upload{
        content_type: "application/octet-stream",
        filename: "currencies.csv",
        path: "/tmp/random_name.csv"
      }
    ]
  ]
}
```
I have only one suggestion how this could be solved - is to implement middleware (as part of absinthe library), which will unwrap list (`List.flatten`). Or we could introduce new `:upload_list` type, which will handle this task properly.

I will appreciate any feedback and help to properly finish this PR. Of course when solution will be good enough I will update documentation.

### DIFF
--- a/lib/absinthe/plug/types.ex
+++ b/lib/absinthe/plug/types.ex
@@ -60,6 +60,12 @@ defmodule Absinthe.Plug.Types do
   """
   scalar :upload do
     parse fn
+      %Blueprint.Input.String{:value => "[]" <> value}, context ->
+        uploads = (context[:__absinthe_plug__][:uploads] || %{})
+        |> Enum.filter(& match_uploads_by_name?(value, &1))
+        |> Enum.map(& elem(&1, 1))
+
+        {:ok, uploads}
       %Blueprint.Input.String{value: value}, context ->
         Map.fetch(context[:__absinthe_plug__][:uploads] || %{}, value)
       %Blueprint.Input.Null{}, _ ->
@@ -71,5 +77,13 @@ defmodule Absinthe.Plug.Types do
     serialize fn _ ->
       raise "The `:upload` scalar cannot be returned!"
     end
+  end
+
+  @spec match_uploads_by_name?(String.t, {String.t, Plug.Upload.t}) :: boolean()
+  defp match_uploads_by_name?(name, {request_upload_name, _}) do
+    request_escaped_name = Regex.escape(name)
+
+    request_upload_name
+    |> String.match?(~r/^#{request_escaped_name}\[\d+\]$/)
   end
 end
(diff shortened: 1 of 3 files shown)


floriank:
I just ran into similar issues when building a small application that should be able to upload files.

I'll try your proposition with [the application](https://github.com/floriank/react_relay_kitten_uploader) as soon as I get around to.

Many thanks for the suggestion <3.

rageofflames:
I really need this... Please review 👍

binaryseed:
Hello there! I reviewed this and I found the convention of `"[]files"` to be difficult to understand.

Is this pattern implemented in any other GraphQL frameworks? Perhaps this problem has already been solved and we can adopt an existing convention instead of creating a new one?

binaryseed:
Since this solution relies on a new GraphQL convention that doesn't exist elsewhere, I don't think we can adopt this.

Perhaps you could submit multiple mutations in a single GraphQL operation, one per file that you wish to upload? This is already possible and part of the GraphQL spec...

Closing this ticket hoping that this solution is workable!

PR #203, opened 2019-05-03
Fix DocumentProvider.pipeline/1 typespec

remiprev:
Hi! Just a small typo fix 😄 

The typespec for the `DocumentProvider.pipeline/1` callback is not the right one. The function does not receive a `%Absinthe.Plug.Request{}` struct — it receives a `%Absinthe.Plug.Request.Query{}` struct.

Thank you!


benwilson512:
Thank you!

PR #204, opened 2019-05-03
Upgrade ex_doc to 0.20.2

maartenvanvliet:
ex_doc 0.20 has many improvements such as full-text search and support for versions


benwilson512:
Thank you!

PR #209, opened 2019-05-28
Fix compiled documents telemetry integration

binaryseed:
This PR makes sure that compiled documents play nicely with `telemetry` integration and ensures they execute instrumentation events like standard documents.

Fix provided by @maartenvanvliet 

Fixes https://github.com/absinthe-graphql/absinthe/issues/734
@benwilson512

### DIFF
--- a/mix.exs
+++ b/mix.exs
@@ -42,7 +42,7 @@ defmodule Absinthe.Plug.Mixfile do
   defp deps do
     [
       {:plug, "~> 1.3.2 or ~> 1.4"},
-      {:absinthe, "~> 1.5.0-alpha.0"},
+      {:absinthe, git: "https://github.com/absinthe-graphql/absinthe.git", branch: "master"},
       {:jason, ">= 0.0.0", only: [:dev, :test]},
       {:ex_doc, "~> 0.20.2", only: :dev}
     ]
--- a/mix.exs
+++ b/mix.exs
@@ -42,7 +42,7 @@ defmodule Absinthe.Plug.Mixfile do
   defp deps do
     [
       {:plug, "~> 1.3.2 or ~> 1.4"},
-      {:absinthe, git: "https://github.com/absinthe-graphql/absinthe.git", branch: "master"},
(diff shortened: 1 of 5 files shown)


mix.exs
```diff
@@ -42,7 +42,7 @@ defmodule Absinthe.Plug.Mixfile do
   defp deps do
     [
       {:plug, "~> 1.3.2 or ~> 1.4"},
-      {:absinthe, "~> 1.5.0-alpha.0"},
+      {:absinthe, git: "https://github.com/absinthe-graphql/absinthe.git", branch: "master"},
```
benwilson512: What is in absinthe master that isn't in the latest alpha release?

test/lib/absinthe/plug/document_provider/compiled_test.exs
```diff
@@ -102,11 +90,47 @@ defmodule Absinthe.Plug.DocumentProvider.CompiledTest do
...
+    assert %Absinthe.Blueprint{} =
+             Compiled.get(Absinthe.Plug.TestLiteralDocuments, "1", :compiled)
   end
 
   test ".get source" do
-    assert @query == Compiled.get(LiteralDocuments, "1", :source)
+    assert @query == Compiled.get(Absinthe.Plug.TestLiteralDocuments, "1", :source)
+  end
+
+  test "telemetry events executed", context do
```
benwilson512: Great test addition to make sure re-adding the phase works.

mix.exs
binaryseed: It's the `telemetry` integration.

We should hold off with this PR until there is another `alpha` that includes telemetry (and i can remove the reference to the git repo)

benwilson512: Ah. We should just go ahead and do that then, I'll push that shortly.

binaryseed: Did a new version get pushed?

binaryseed: todo: update this now that a new release of absinthe is out

PR #217, opened 2019-08-31
Add support for brokered subscriptions

jscheid:
This is work in progress, for now I'm putting it up here only to start the conversation. This change goes along with https://github.com/absinthe-graphql/absinthe/pull/776

Like the changes in the other PR, this one is necessary to support message brokers for delivering subscription updates because obviously, updates aren't being sent through the WebSocket in that case (they are sent out-of-band through the broker instead.)

One question is how to encode the subscription ID in the HTTP response. GraphQL Ruby delivers the string `null` in the body with Content-Type `application/json` and sets the subscription ID in the `X-Subscription-ID` header.  At first I wanted to do it the same way for consistency, but doing so doesn't work well with one of my clients, Apollo iOS.  It chokes on the `null` (it wants a JSON body that is an object or an array), and more importantly it doesn't provide a good way to access the GraphQL response headers.

As far as I know, there is no standard for this and the GraphQL Ruby guys just made something up (?).  So unless there are any objections I'd prefer to just put the ID in the body instead.

(I've also come to think that it's slightly more consistent because the body is where any errors would come in, so by putting the ID in the body you don't have to look in two different places.)

Thoughts?

### DIFF
--- a/lib/absinthe/plug.ex
+++ b/lib/absinthe/plug.ex
@@ -142,6 +142,7 @@ defmodule Absinthe.Plug do
   - `:log_level` -- (Optional) Set the logger level for Absinthe Logger. Defaults to `:debug`.
   - `:analyze_complexity` -- (Optional) Set whether to calculate the complexity of incoming GraphQL queries.
   - `:max_complexity` -- (Optional) Set the maximum allowed complexity of the GraphQL query. If a document’s calculated complexity exceeds the maximum, resolution will be skipped and an error will be returned in the result detailing the calculated and maximum complexities.
+  - `:websocket_subscriptions` -- (Optional) Whether to assume subscriptions are sent through and receive updates from a WebSocket. (default: `true`).
 
   """
   @type opts :: [
@@ -160,7 +161,8 @@ defmodule Absinthe.Plug do
           serializer: module | {module, Keyword.t()},
           content_type: String.t(),
           before_send: {module, atom},
-          log_level: Logger.level()
+          log_level: Logger.level(),
+          websocket_subscriptions: boolean
         ]
 
   @doc """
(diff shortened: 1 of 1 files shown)


lib/absinthe/plug.ex
```diff
@@ -301,10 +306,15 @@ defmodule Absinthe.Plug do
   def subscribe(conn, topic, %{context: %{pubsub: pubsub}} = config) do
     pubsub.subscribe(topic)
 
-    conn
-    |> put_resp_header("content-type", "text/event-stream")
-    |> send_chunked(200)
-    |> subscribe_loop(topic, config)
+    if config[:websocket_subscriptions] do
```
benwilson512: This is the wrong option name I think. What Absinthe.Plug is doing here is NOT websockets, it's long polling. Websockets is handled by absinthe phoenix, and happens when a subscription is sent over a pre-established websocket channel. Everything in Absinthe.Plug is HTTP.

However I get what you're going for. You want to make it possible to simply return a subscription handle, and indicate that the subscription was registered elsewhere. This is a need we'll honestly have for both Absinthe.Plug and Absinthe.Phoenix, cause in theory nothing should stop you from submitting a subscription document over websocket, but having that still set something up in MQTT.

However, I think this needs to be possible on a request by request basis, instead of as a flag that categorically sets a given plug one way or another. Not entirely sure how that option would be specified however. It seems like it could be something that Absinthe itself indicates in the return value from a subscribe call.

jscheid: Ok, sure, that would work too. How should we do this -- do you want to think more about it and let me know how you want to structure it? Or were you planning to build something different yourself?

binaryseed:
Closing this since it's grown stale and there isn't a clear path forward. Please update us if anything changes!

Issue #199, opened 2019-02-18
GraphiQL hangs at Loading GraphQL Playground


xshyne88:
Hi there,
```
   scope "/graphiql" do
    pipe_through([:api, :public, :absinthe])

    forward(
      "/",
      Absinthe.Plug.GraphiQL,
      schema: docsWeb.Schema,
      json_codec: Jason,
      interface: :playground,
      socket: docsWeb.UserSocket,
      pipeline: {ApolloTracing.Pipeline, :plug}
    )
```
   this is our router.ex, when we comment out interface and not use the graph-cool interface the playground loads just fine, otherwise it hangs on loading graphql playground and this is the javascript error we get

```
Uncaught SyntaxError: Unexpected end of input
graphiql:69 Uncaught ReferenceError: GraphQLPlayground is not defined
    at graphiql:69
```

 (I should note these error messages differ accross browsers but all have to do with graphiql)

Our team tried this across different browsers, machines and os's. We originally thought it was a browser cacheing issue where the middleware CDN had only served part of the file, and our browsers were cacheing it, but we've cleared, disabled, and force refreshed and still no luck.
Apologies in advanced if this isn't absinthe plug related though I'm fairly certain it is because we aren't having issues loading the graphcool interface on other services.

If we comment out the ```interface: :playground``` and load the non graphcool version, it loads just fine, which is also weird. 
Thanks in advance for any help!!

Side note: these absinthe libraries are absolutely excellent and we've had almost no problems up to this point!

maartenvanvliet:
We also hit this since today. Still investigating the what's causing it.

bruce:
I can reproduce this as well, on a project that `absinthe_plug` hasn't been recently updated and after updating `absinthe_plug` to latest (and on `master`):

I'm seeing the following:

```
SyntaxError: missing } after function body[Learn More] middleware.js:1:343596 note: { opened at line 1, column 343516 (middleware.js:1:343516)
ReferenceError: GraphQLPlayground is not defined[Learn More]
```

I'm going to mention this over at https://github.com/absinthe-graphql/absinthe-graphql-playground, in the event @tlvenn or @mgtitimoli have any ideas what might be going on.

fredr:
I tried changing to the `:playground`-interface in our project running `absinthe_plug 1.4.6`, but I can't reproduce the problem.

Maybe there is something weird going on with the CDN delivering the js.

Does your middleware.js look broken? 
[...]

PR #222, opened 2020-01-11
Add Altair GraphQL interface.

imolorhe:
This PR introduces the [Altair GraphQL](https://altair.sirmuel.design/) interface, as an alternative to GraphiQL and GraphQL playground.

Altair is a more fully featured alternative to the other options and has become a household name for working with GraphQL. It also seems to not suffer from issues affecting other tools (e.g. #199).

Please review and let me know what you think! 😊

### DIFF
--- a/lib/absinthe/plug/graphiql.ex
+++ b/lib/absinthe/plug/graphiql.ex
@@ -25,6 +25,15 @@ defmodule Absinthe.Plug.GraphiQL do
           interface: :simple
         ]
 
+  Use the "altair" interface for Altair GraphQL instead:
+
+      forward "/graphiql",
+        to: Absinthe.Plug.GraphiQL,
+        init_opts: [
+          schema: MyAppWeb.Schema,
+          interface: :altair
+        ]
+
   Finally there is also support for GraphiQL Playground
   https://github.com/graphcool/graphql-playground
 
@@ -43,8 +52,9 @@ defmodule Absinthe.Plug.GraphiQL do
   - `:advanced` (default) will serve the [GraphiQL Workspace](https://github.com/OlegIlyenko/graphiql-workspace) interface from Oleg Ilyenko.
   - `:simple` will serve the original [GraphiQL](https://github.com/graphql/graphiql) interface from Facebook.
   - `:playground` will serve the [GraphQL Playground](https://github.com/graphcool/graphql-playground) interface from Graphcool.
+  - `:altair` will serve the [Altair GraphQL](https://altair.sirmuel.design/) interface.
 
-  See `Absinthe.Plug` for the other  options.
+  See `Absinthe.Plug` for the other options.
 
   ## Default Headers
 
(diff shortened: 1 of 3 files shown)


smaximov:
Thanks for the PR, @imolorhe, I looked up Altair and it looks great! I also like that it's apparently more maintained than the alternatives.

I have a small request: can you also add support for [`:default_headers`](https://hexdocs.pm/absinthe_plug/1.5.0-rc.0/Absinthe.Plug.GraphiQL.html#module-default-headers) similar to how it is done for GraphiQL Workspace? According to the [README](https://github.com/imolorhe/altair#configuration-options), Altair supports it as the `initialHeaders` configuration option.

imolorhe:
@smaximov I wouldn't mind adding it. However, I'm not conversant enough with elixir to know the right way to pass a dictionary type (GraphiQL seems to use headers as an array, but Altair uses a dictionary/map/object type). If you can give me a pointer in the right direction, I wouldn't mind adding that. 🙂

smaximov:
@imolorhe I believe you should just encode it as JSON. This is done in `handle_default_headers` for GraphiQL Workspace:

https://github.com/absinthe-graphql/absinthe_plug/blob/fc938660a393c1b5f9c4eef46a83dfff3b19d1e0/lib/absinthe/plug/graphiql.ex#L407-L411

The trick is to select the correct headers representation depending on the configured interface. Something like this should do:

``` elixir
  defp handle_default_headers(%{interface: interface} = config, conn) do
    default_headers =
      case get_config_val(config, :default_headers, conn) do
        nil ->
          {}

        val when is_map(val) ->
          val

        val ->
          raise "invalid default headers: #{inspect(val)}"
      end

    header_string =
      default_headers
      |> prepare_default_headers(interface)
      |> config.json_codec.module.encode!(pretty: true)

    Map.put(config, :default_headers, header_string)
  end
  
  # headers will be encoded as an array of objects (for Playground)
  defp prepare_default_headers(headers, :advanced),
    do: Enum.map(headers, fn {k, v} -> %{"name" => k, "value" => v} end)

  # headers will be encoded as an object (for Altair)
  defp prepare_default_headers(headers, :altair), do: headers

  # fail if unsupported interface is configured
  defp prepare_default_headers(_headers, interface),
    do: raise(":default_headers option not supported for interface #{inspect(interface)}")
```

Please tell me if you need help with it.

benwilson512:
Hi @imolorhe thanks for the PR. It is my sense that adding additional GraphiQL UIs is not the way to go, in fact in a future version of `absinthe_plug` I'd prefer to include NO GraphiQL interfaces (or maybe just the "standard" one) and instead have each different interface provided as a distinct package.

There are dozens of GraphiQL like clients out there and it just isn't feasible to have the core team maintain all of them. I don't see anything here in this PR that couldn't be provided by an external library and maintained by those who care about it most. @bruce thoughts?

bruce:
@benwilson512 I agree. Altair is my favorite GraphQL client, but as we’ve discovered with our current packaged clients, it is very difficult to stay up-to-date with releases and manage issues with CDNs over time while maintaining our focus on Absinthe itself.

@imolorhe Please don’t be discouraged by our resistance to including this. I think this idea would make a great third-party package that we’d be happy to link to in docs. The decision really has more to do with us trying to keep Absinthe’s project scope focused—it’s already quite a bit for us maintainers to take care of and improve.

imolorhe:
Alright then

imolorhe:
I can't seem to close the PR. Feel free to close it.

PR #230, opened 2020-05-25
Add helper to insert values into context

jayjun:
`Absinthe.Plug.put_options(conn, context: context)` overwrites any previously set context.

This pull request adds `assign_context` which merges values instead. This allows multiple plugs that configure context to be composed.

```elixir
conn
|> Absinthe.Plug.assign_context(current_user: current_user)
|> Absinthe.Plug.assign_context(foo: "bar")
```

I considered naming it `put_in_context` but conventionally `put` and `put_in` functions do not merge. Phoenix’s `assign` appears closest in behaviour.

### DIFF
--- a/lib/absinthe/plug.ex
+++ b/lib/absinthe/plug.ex
@@ -353,6 +353,36 @@ defmodule Absinthe.Plug do
     Plug.Conn.put_private(conn, :absinthe, Enum.into(opts, %{}))
   end
 
+  @doc """
+  Adds key-value pairs into Absinthe context.
+
+  ## Examples
+
+      iex> Absinthe.Plug.assign_context(conn, current_user: user)
+      %Plug.Conn{}
+  """
+  @spec assign_context(Plug.Conn.t(), Keyword.t() | map) :: Plug.Conn.t()
+  def assign_context(%Plug.Conn{private: %{absinthe: absinthe}} = conn, assigns) do
+    context =
+      absinthe
+      |> Map.get(:context, %{})
+      |> Map.merge(Map.new(assigns))
+
+    put_options(conn, context: context)
+  end
+
+  def assign_context(conn, assigns) do
+    put_options(conn, context: Map.new(assigns))
+  end
+
+  @doc """
+  Same as `assign_context/2` except one key-value pair is assigned.
+  """
+  @spec assign_context(Plug.Conn.t(), atom, any) :: Plug.Conn.t()
+  def assign_context(conn, key, value) do
+    assign_context(conn, [{key, value}])
+  end
+
   @doc false
   @spec execute(Plug.Conn.t(), map) :: {Plug.Conn.t(), any}
   def execute(conn, config) do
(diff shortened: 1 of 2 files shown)


lib/absinthe/plug.ex
```diff
@@ -353,6 +353,36 @@ defmodule Absinthe.Plug do
...
+
+      iex> Absinthe.Plug.assign_context(conn, current_user: user)
+      %Plug.Conn{}
+  """
+  @spec assign_context(Plug.Conn.t(), Keyword.t() | map) :: Plug.Conn.t()
+  def assign_context(%Plug.Conn{private: %{absinthe: absinthe}} = conn, assigns) do
+    context =
+      absinthe
+      |> Map.get(:context, %{})
+      |> Map.merge(Enum.into(assigns, %{}))
```
benwilson512, line 369: The merge order here is probably counter to what you want. `assigns` should be merged into the current context not the other way around, so that if you _do_ want to override an existing assign it works as expected.

```suggestion
    context =
      assigns
      |> Map.new
      |> Map.merge(Map.get(absinthe, :assigns, %{}))
```

lib/absinthe/plug.ex
```diff
@@ -353,6 +353,36 @@ defmodule Absinthe.Plug do
...
+
+    put_options(conn, context: context)
+  end
+
+  def assign_context(conn, assigns) do
+    put_options(conn, context: Enum.into(assigns, %{}))
```
benwilson512, line 375:
```suggestion
    put_options(conn, context: Map.new(assigns))
```

test/lib/absinthe/plug_test.exs
```diff
@@ -514,6 +514,36 @@ defmodule Absinthe.PlugTest do
...
+    test "values can be added individually" do
+      conn =
+        conn(:post, "/")
+        |> Absinthe.Plug.assign_context(:current_user, %{id: 1})
+        |> Absinthe.Plug.assign_context(:foo, "bar")
+
+      assert conn.private.absinthe.context.current_user.id == 1
+      assert conn.private.absinthe.context.foo == "bar"
+    end
+  end
```
benwilson512, line 545: Can you add a test case that tests assigning a value that is already assigned?

lib/absinthe/plug.ex
jayjun: My understanding is that right wins.
```elixir
iex> Map.merge(%{foo: "bar"}, %{foo: "baz"})      
%{foo: "baz"}
```
So the current code un-pipelined
```elixir
context = Map.get(absinthe, :context, %{})
context = Map.merge(context, Map.new(assigns))
```
looks correct to me.

lib/absinthe/plug.ex
jayjun: Done! `put_options` itself needs the same change. Let me know if you want those included here.

test/lib/absinthe/plug_test.exs
jayjun: Done.

lib/absinthe/plug.ex
benwilson512: You are correct! I got Map.merge and Enum.into backwards, my bad!

jeroenvisser101:
@benwilson512 are you planning to released this in 1.5.1(-rc)? We'd love to use this :)

PR #231, opened 2020-06-20
upgrade the different graphql web interfaces and dependencies

mayel:


lib/absinthe/plug/graphiql/graphiql_playground.html.eex
```diff
@@ -55,7 +56,19 @@ add "&raw" to the end of the URL within a browser.
     </div>
   </div>
   <script>
-    var options = {};
+    var options = { // FIXME: looks like these options are ignored
```
benwilson512, line 59: Can you elaborate on this?

lib/absinthe/plug/graphiql/assets.ex
```diff
@@ -35,38 +45,46 @@ defmodule Absinthe.Plug.GraphiQL.Assets do
...
-    {"@absinthe/graphql-playground", "1.2.0",
-     [
-       {"build/static/css/middleware.css", "playground.css"},
-       {"build/static/js/middleware.js", "playground.js"}
-     ]},
-    {"@absinthe/socket-graphiql", "0.1.1",
+    {
+      "graphql-playground-react",
+      "1.7.23",
+      # should probably be deprecated eventually, as playground is being shut down in favour of GraphiQL: https://github.com/prisma-labs/graphql-playground/issues/1143
```
benwilson512, line 81: I think we should honestly deprecate these now and begin the process of extracting _all_ of the graphiql interfaces into dedicated hex deps. The plan with Absinthe.Plug 2.0 will be to not have any built in.

mayel: You plan sounds good. Some people might want a Playground hex option then. Workspace on the other hand is unmaintained but very useful (it makes it easy to add custom headers). So until the next major version of GraphiQL comes out (with pluginisation) both of these are handy...

benwilson512: For sure.

benwilson512:
Have you tested these updates with subscriptions by the way?

mayel:
Ah no, as I'm not yet using subscriptions

binaryseed:
Closing this particular ticket since our path forward is to remove GraphiQL from `Absinthe.Plug`...

PR #232, opened 2020-07-30
Compile queries from files

### LINKED ISSUE #140: Be able to compile queries loaded from textfile
The current Compiled DocumentProvider macro does not work if you want to provide queries from files and set the key in the Document Provider. 

I would like to be able to use the provide macro like this: 

```
provide %{
    "key" -> File.read!("query.txt")
}
```

vgsantoniazzi:
This closes #140, where @sweco-semtne referenced a bug that
do not occurs anymore.


vgsantoniazzi:
@benwilson512 What do you think?

PR #234, opened 2020-10-31
Alternative schema validation check

binaryseed:
When a schema is using `persistent_term` backend, asking for the `types` during compile time will cause an exception since they haven't been populated yet.

This PR provides an alternative validation check that verifies if the atom is a module and if that module implements the `Absinthe.Schema` behavior.


benwilson512 (approved):
Definitely a better way to check!

PR #236, opened 2020-11-05
Fix Elixir 1.11 compiler warnings

rzane:
See: https://elixirforum.com/t/how-to-fix-elixirs-1-11-application-boundaries/35383/7


benwilson512:
Thanks!

PR #237, opened 2020-11-26
Transport Batch payload configuration option

binaryseed:
This PR adds a configuration option that enables opting out of the default behavior for Transport Batch requests which nests the result in a `payload` key.

Followup to #179 from @aleqsio


binaryseed:
Closes #169

PR #243, opened 2021-01-18
Set all plug options via put_options

### LINKED ISSUE #241: Dynamically generated max complexity
Hello! 👋 First of all, thank you for Absinthe and `Absinthe.Plug`!

I'm wondering if the maintainers would be open to the addition of a feature to dynamically determine the max complexity per query. I'd like to be able to say something like, "User X made a query, they have Y complexity allowance left, will I allow them to make this query?"

Of course this doesn't match the primary use case. 😄 

Whereas the current API looks like this:

```elixir
plug Absinthe.Plug,
  schema: MyAppWeb.Schema,
  analyze_complexity: true,
  max_complexity: 50
```

I would like to propose expanding the API to support this as well:

```elixir
plug Absinthe.Plug,
  schema: MyAppWeb.Schema,
  analyze_complexity: true,
  max_complexity: &MyModule.max_complexity/1
```

Where `&MyModule.max_complexity/1` is a function that is given the context and query and returns back an integer.

Thank you for considering! I would be glad to whip up a PR if you're open to this. Cheers!

binaryseed:
This PR wires up the ability to set complexity settings via `put_options`

@benwilson512 These make sense to set on a per-request basis, but I'm not sure if the other options do...

closes #241

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Changelog
 
+## v1.5.4
+
+- Feature: [Set all plug options via put_options](https://github.com/absinthe-graphql/absinthe_plug/pull/243)
+
 ## v1.5.3
 
 - Chore: [Loosen Absinthe version](https://github.com/absinthe-graphql/absinthe_plug/pull/242)
--- a/lib/absinthe/plug.ex
+++ b/lib/absinthe/plug.ex
@@ -124,7 +124,27 @@ defmodule Absinthe.Plug do
 
   alias __MODULE__.Request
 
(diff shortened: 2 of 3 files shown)


benwilson512:
@binaryseed Looking through the options list, I could imagine almost all of them being relevant based on different headers or something. Adapters, whether you have before_send, I think literally all of them could be relevantly different based on client header or path values.

binaryseed:
Ok, wired it up, mind giving it a quick review?

benwilson512 (approved):
Looks great!

binaryseed:
Released in `1.5.4`

PR #244, opened 2021-01-18
Revert incorrect readme changes

dylan-chong:
Wasn't supposed to change that. oops


binaryseed:
I think this change is fine!

PR #245, opened 2021-01-18
updated project list link

bmuller:
Not sure where the project list link should go to, but the current location results in a 404.


binaryseed:
Thanks!

Issue #247, opened 2021-02-01
upgrade from 1.5.3 to 1.5.4 breaks default pub sub option


woylie:
The upgrade from 1.5.3 to 1.5.4 broke our application. We didn't set the `pubsub` option explicitly, but relied on the default `conn.private[:phoenix_endpoint]`. After the upgrade, this default apparently gets overwritten by the `update_config` function. We noticed because subscription triggers still worked when pushing mutations via a socket, but they didn't work anymore when sending a mutation via HTTP.

benwilson512:
Hi @woylie, are you setting `pubsub` explicitly in update_config? It shouldn't be overriding it with `nil`, can you elaborate?

woylie:
We're not using `update_config` at all.

binaryseed:
Here's the PR that recently changed this logic: https://github.com/absinthe-graphql/absinthe_plug/pull/243

Though quickly looking through it I don't see anything obviously wrong...

benwilson512:
Right, @woylie how / where would update_config override things? How are you determining that pubsub is being unset?

woylie:
I'm sorry, I didn't have any time to investigate this any further. I don't actually know that `pubsub` is being unset, that was my assumption based on:

- Upgrading from 1.5.3 to 1.5.4 causes mutations sent via HTTP not triggering any subscription messages, while mutations via socket still trigger them as expected.
- This problem can be reliably fixed/reproduced by down/upgrading between 1.5.3 and 1.5.4. No other changes are necessary, no other dependency versions are changed in `mix.lock`.
- The only change between 1.5.3 and 1.5.4 is in the `update_config` function.

woylie:
This happens with absinthe 1.5.5 and absinthe_phoenix 2.0.0.

benwilson512:
So the actual observed behavior is this:
> mutations sent via HTTP not triggering any subscription messages, while mutations via socket still trigger them as expected.

woylie:
Yes. Sorry for not being clearer in the first description.

binaryseed:
We figured it out, here's a fix: https://github.com/absinthe-graphql/absinthe_plug/pull/249

woylie:
Great, thanks!

PR #248, opened 2021-02-04
Prevent config updates from overwriting map values

ejscunha:
Map values coming from init_options were overwriting the previous value instead of merging both maps.

This was removing the pubsub key from the config context map, which made subscription's triggers from mutations stop working.

I have no idea on how to test this fix.

### DIFF
--- a/lib/absinthe/plug.ex
+++ b/lib/absinthe/plug.ex
@@ -343,11 +343,23 @@ defmodule Absinthe.Plug do
   end
 
   defp update_config(config, :init_options, %{private: %{absinthe: absinthe}}) do
-    Map.merge(config, Map.take(absinthe, @init_options -- @raw_options))
+    absinthe_config = Map.take(absinthe, @init_options -- @raw_options)
+    merge_config(config, absinthe_config)
   end
 
   defp update_config(config, _, _conn) do
     config
+  end
+
+  defp merge_config(config, config_to_merge) do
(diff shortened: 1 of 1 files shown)


benwilson512:
A test would really help me understand what the issue is, can you at least supply some examples from your own code and then I can try to write a test?

binaryseed:
Will it also work to just swap the order of the `update_config` so that the `pubsub` one goes last?

benwilson512:
Oh I see, the issue is that if you set `context: %{foo: :bar}`, it nukes the context pubsub value that is set. @binaryseed I like your solution.

eduardo-cunha-bose:
> A test would really help me understand what the issue is, can you at least supply some examples from your own code and then I can try to write a test?

I will try to setup a test tomorrow for this.

> Will it also work to just swap the order of the update_config so that the pubsub one goes last?

If we change the order we lose the added context information.

eduardo-cunha-bose:
Looking into the issues, #247 is what I'm observing and this PR should fix it (it did on my project).

binaryseed:
Here's a fix that works by re-ordering functions in `update_config` https://github.com/absinthe-graphql/absinthe_plug/pull/249 & a test for it

Thanks for the PR anyway!

eduardo-cunha-bose:
Oh that makes sense, I didn't understand at first. Thank you for taking caring of that quickly. 🙏

PR #249, opened 2021-02-04
Fix update_config bug

### LINKED ISSUE #247: upgrade from 1.5.3 to 1.5.4 breaks default pub sub option

binaryseed:
This PR fixes a bug in the `update_config` function so that it doesn't stomp any existing values in the `context`

Fixes #247

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Changelog
 
+## v1.5.5
+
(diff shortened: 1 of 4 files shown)


benwilson512:
Thanks for handling this!

PR #253, opened 2021-03-22
Improve Github Actions

dolfinus:
Hello.

It looks like commiting to branches excluding master does not run Github Actions. So users which are working on feature/bugfix branches cannot see that their changes are leading to failed tests until they'll create a pull request to master branch.
So I've removed restriction against branch name for push events.

Also I've found another quite annoying bug - step actions/cache does not update cache item if it was restored using a primary key.
For example, some dependencies like dialyzer or other ones store their files in deps or _build folders (which are cached in the workflow). But when these dependencies perform some updates of these folders, it does not lead to the updating of mix.lock file. File hasn't been changed -> hash is the same -> cache primary key is the same -> caching action will not update the existing cache.
This will cause the increase of workflow run time because the stored cache will not be used by some steps. Until mix.lock will be updated which will lead to updating the cache item.
Here I've appended a commit SHA to the cache primary key, so it will always be updated. But it will not affect cache reading.

What do you think about that?

### DIFF
--- a/.github/workflows/elixir.yml
+++ b/.github/workflows/elixir.yml
@@ -2,8 +2,6 @@ name: CI
 
 on:
   push:
-    branches:
-      - master
   pull_request:
     branches:
       - master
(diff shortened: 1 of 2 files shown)


dolfinus:
This PR is just the same as already merged https://github.com/absinthe-graphql/absinthe_relay/pull/181.

binaryseed:
Thanks!

PR #254, opened 2021-03-23
Update README

dolfinus:
Also let's sync README files with other projects, e.g. https://github.com/absinthe-graphql/absinthe


benwilson512:
Thank you!

PR #256, opened 2021-04-08
Fix bug with over encoded JSON body

binaryseed:
This PR fixes an edge case handling an over encoded JSON string in the POST body


benwilson512 (approved):
Ah great catch.

binaryseed:
Thanks to our customers who always find creative ways of sending API requests :)

PR #267, opened 2022-05-04
Update Absinthe GraphiQL example docs

vitor-l:
Hello!

I tried to use the examples in the current doc, but It did not work. So I'm opening this pull request with the solution I found for my problem. I hope that helps.


benwilson512:
Were you using a phoenix router or a plug router?

vitor-l:
Phoenix router

PR #276, opened 2022-11-08
Switch to Application.compile_env/3

srfsh:
It gets rid of this warning:

```
==> absinthe_plug
Compiling 18 files (.ex)
warning: Application.get_env/2 is discouraged in the module body, use Application.compile_env/3 instead
  lib/absinthe/plug/graphiql/assets.ex:5: Absinthe.Plug.GraphiQL.Assets
```
  
Since we are requiring `~> 1.10` version of Elixir, this change should be alright.


benwilson512:
Thanks!

PR #278, opened 2023-01-30
Rescue exceptions, raise Plug.Conn.WrapperError

vincedevendra:
We have a Phoenix application that expects the `[:phoenix, :endpoint, :stop]` event to be executed, but I noticed that this was not happening when uncaught exceptions originated from this plug.

This is because `Plug.Telemetry` [registers this event](https://github.com/elixir-plug/plug/blob/63ddce30077d58470215f2aba9eb3ca9f7416140/lib/plug/telemetry.ex#L74) by using `Plug.Conn.register_before_send`, which, in turn, adds a key to `conn.private`.

When there is an uncaught exception, Phoenix catches the error to render a standard error response [here](https://github.com/phoenixframework/phoenix/blob/53d36bc63013d9a84040bf811ff01c07c8c7e040/lib/phoenix/endpoint/render_errors.ex#L38).

Notice that if there is a `Plug.Conn.WrapperError`, the `conn` from it is used, but if not, `conn` is what's in scope at the time the `RenderErrors` plug runs. This happens to be at the top of the middleware stack before `Plug.Telemetry` registers the callback. Thus, no `[:phoenix, :endpoint, :stop]` event is executed. 

Note also that bug reporting plugs may also expect request parsers etc. to have surfaced important bits of info on the `conn` (we ran into this issue as well, but worked around it).

This issue can be solved by rescuing errors in `call` and using `Plug.Conn.Wrapper.reraise/4`

### DIFF
--- a/lib/absinthe/plug.ex
+++ b/lib/absinthe/plug.ex
@@ -286,37 +286,43 @@ defmodule Absinthe.Plug do
   """
   @spec call(Plug.Conn.t(), map) :: Plug.Conn.t() | no_return
   def call(conn, config) do
-    config = update_config(conn, config)
-    {conn, result} = conn |> execute(config)
+    try do
+      config = update_config(conn, config)
+      {conn, result} = conn |> execute(config)
 
-    case result do
-      {:input_error, msg} ->
-        conn
-        |> encode(400, error_result(msg), config)
+      case result do
+        {:input_error, msg} ->
+          conn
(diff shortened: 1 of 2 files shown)


vincedevendra:
I should note that this likely only effects Phoenix applications _without_ a router, since Phoenix Router wraps errors. As our application is solely a GraphQL server, we don't have/need a router.

PR #279, opened 2023-02-28
Better errors for schema module validation failures

grantwest:
Addresses 2 schema module validation issues:

Use `Code.ensure_compiled!/1` rather than `Code.ensure_compiled/1` to avoid masking compile issues and avoid potential race conditions. The existing pattern is specifically warned against in the elixir docs: https://hexdocs.pm/elixir/1.14/Code.html#module-code-loading-on-the-erlang-vm

With the current code it is possible to get the`is not valid Absinthe.Schema` error even though the issue is actually a failure to compile another module. This creates a painful experience.

The second improvement is to specify exactly why a schema has failed validation. If a validation fails, the next question asked is why? This change answers that question immediately, rather than making the user change absinthe_plug code to debug.

Because `Code.ensure_compiled!/1` is introduced in elixir 1.12 this commit also bumps the minimum elixir version up to 1.12.


benwilson512:
I am open to bumping the Elixir requirement to 1.12.0

grantwest:
Ok, I bumped the min elixir version in this same commit to maintain the reason for the bump in the commit history.

grantwest:
I refreshed this PR.

cschiewek (approved):
Thank you!

PR #280, opened 2023-03-01
Update github workflow to match plug

grantwest:
The workflow was failing on not being able to install the specified erlang version. This change updates the workflow config to match https://github.com/elixir-plug/plug


benwilson512:
Thanks!

PR #281, opened 2023-03-08
Add token limit to init and raw options

griffitr:
Hello! This is a small amount of follow-up work to my teammate's (@stellanor) addition of token limiting in Absinthe. Here is that prior context: https://github.com/absinthe-graphql/absinthe/pull/1210

This work adds the new `token_limit` option so that we can leverage the new ability in Absinthe. I have tested this locally with our own service against the current master branch of Absinthe to verify the value is passed along as expected, but as of opening this PR, I haven't added an integration test for this; I believe we'd need to upgrade the Absinthe dependency to the current master branch (so that the effects of token limiting are present) in order to do so. Then again, the value does behave the same way/follow the same flow as `max_complexity`, so I'm also wondering if a dedicated test is necessary. The token limiting feature itself is fully tested in Absinthe.

I'm happy to receive any feedback/requests to get functionality this added in, thank you!

### DIFF
--- a/lib/absinthe/plug.ex
+++ b/lib/absinthe/plug.ex
@@ -139,11 +139,13 @@ defmodule Absinthe.Plug do
     :pubsub,
     :analyze_complexity,
     :max_complexity,
+    :token_limit,
     :transport_batch_payload_key
   ]
   @raw_options [
     :analyze_complexity,
-    :max_complexity
+    :max_complexity,
+    :token_limit
   ]
 
   @type function_name :: atom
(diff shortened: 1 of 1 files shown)


mattbaker:
@benwilson512 any chance you'd be able to look at this? We're eager to get access to https://github.com/absinthe-graphql/absinthe/pull/1210 😄 

We're not opposed to running a fork (I know you're busy) but it'd be nice to get this to a point where it can be merged if you have a chance to give it a look.

benwilson512:
Apologies I simply missed this!

mattbaker:
Woo! Thanks so much Ben! 🎉

griffitr:
Hoorayyyyyy!!!! ✨

sebastianseilund:
@benwilson512 Could we get a new `absinthe_plug` release so we can use `token_limit` in `Absinthe.Plug`? Thanks a lot :)

PR #282, opened 2023-04-10
Added Apollo's Explorer Sandbox interface

rwillians:
Playground project has been [deprecated in Dec 2022](https://www.apollographql.com/docs/apollo-server/v2/testing/graphql-playground/) giving space to [Apollo's Explorer embedded Sandbox](https://www.apollographql.com/docs/graphos/explorer/sandbox#embedding-sandbox) (free -- no account required).

In this PR:
- Added support for Apollo's Explorer Sandbox interface (`:apollo_explorer_sandbox`) and updated docs;
- Minor refac on `Absinthe.Plug.GraphiQL`'s private functions;
- Bumped `absinthe_plug` version from `1.5.8` to `1.6.0`.

### DIFF
--- a/config/config.exs
+++ b/config/config.exs
@@ -1,5 +1,9 @@
 # This file is responsible for configuring your application
 # and its dependencies with the aid of the Mix.Config module.
+
+# `Mix.Config` is deprecated since elixir's version v1.12.3.
+# @TODO: Replace `use Mix.Config` for `import Config` on v2 release and bump
+#        Elixir's minimum version to v1.12.3.
 use Mix.Config
 
 # This configuration is loaded before any dependency and is restricted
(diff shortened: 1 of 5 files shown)


rwillians:
Apollo Explorer Sandbox isn't working locally for subscriptions -- using websockets in general.

The root cause seems to be associated to the fact that the `Origin` in the upgrade request is an Apollo hostname and, since the local environment isn't server using SSL (who uses SSL locally??), the browser (chromium-based in my case) aborts the request before it even hits the server.

No to mention the Sandbox tracks a lot of data, which isn't nice.

For those reasons, I'm closing my PR.
For whoever interests in using Apollo Sandbox despite of those reason, you can use it from https://studio.apollographql.com/sandbox/explorer/

Issue #284, opened 2023-07-07
Support spec_compliant_errors in Plug options


thecynicalpaul:
As per https://github.com/absinthe-graphql/absinthe/pull/1215, it'd be great to be able to support the modern GraphQL spec for errors while using this Plug (https://spec.graphql.org/October2021/#sec-Errors.Error-result-format).

benwilson512:
PR welcome!

PR #285, opened 2023-07-07
Feat/add spec compliant errors opt

thecynicalpaul:
As per the spec (https://spec.graphql.org/October2021/#sec-Errors.Error-result-format) enables placing extra errors in the extensions field using `spec_compliant_errors: bool` plug option.

In line with https://github.com/absinthe-graphql/absinthe/pull/1215, this is an optional prop with a default of `false` until there is a major version release.

Fixes: #284

### DIFF
--- a/lib/absinthe/plug.ex
+++ b/lib/absinthe/plug.ex
@@ -575,7 +575,7 @@ defmodule Absinthe.Plug do
     config.schema_mod
     |> Absinthe.Pipeline.for_document(pipeline_opts)
     # See: https://github.com/absinthe-graphql/absinthe/issues/925
-    |> Absinthe.Pipeline.replace(Absinthe.Phase.Document.Result, {Absinthe.Phase.Document.Result, merged_opts})
+    |> Absinthe.Pipeline.replace(Absinthe.Phase.Document.Result, {Absinthe.Phase.Document.Result, merged_pipeline_opts})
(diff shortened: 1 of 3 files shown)


benwilson512:
Hey @thecynicalpaul I think this is a good contribution, but I am wondering if there is a more generic way to handle this so that it doesn't need to be managed this way with a custom pipeline. This limits the ease of use of this feature.

thecynicalpaul:
> Hey @thecynicalpaul I think this is a good contribution, but I am wondering if there is a more generic way to handle this so that it doesn't need to be managed this way with a custom pipeline. This limits the ease of use of this feature.

@benwilson512 Thanks! I am not 100% sure. I think it really depends on what is going to happen to Absinthe core. Right now, the `spec_compliant_errors` option is only supplied to the result phase. There isn't a simple way to use this for the end-user at the moment. Do you have any thoughts?

kradydal:
Are you planning to merge it? It would be great to have this option in the absinthe_plug 🙏

maartenvanvliet:
A PR to the absinthe core document pipeline where the options are passed through is first needed. Then the option can be added to absinthe_plug as well.

benwilson512:
Hey folks! I'm trying to follow up on the status of some of these PRs. Did the PR mentioned by @maartenvanvliet get submitted and merged?

PR #292, opened 2024-07-29
Change SSE response structure so that it's compatible with GraphQL SSE standard

perzanko:
**Description**
The current payload of a subscription response in the SSE protocol looks like this.

```
{"data": {."foo": "bar" } }\n\n
```
It is simply encoded JSON with a new line.

The proposed response format is as follows:
```
event: next\ndata: {"data": {."foo": "bar" } }\n\n
```
The encoded response additionally contains an event segment `event: next` and the data is prefixed with a `data: ` field, indicating data segment. This structure is consistent with [GraphQL over Server-Sent Events Protocol specification](https://github.com/enisdenjo/graphql-sse/blob/master/PROTOCOL.md#next-event) and official [SSE standard](https://html.spec.whatwg.org/multipage/server-sent-events.html#event-stream-interpretation).

**Reason**
Many graphql clients expect a response consistent with the above standard. An example of this is [graphql-mesh](https://the-guild.dev/graphql/mesh/v1/serve/features/subscriptions#example), which expects responses from subgraph services sent in a structured manner. Or even the highly popular client [graphql-sse](https://www.npmjs.com/package/@graphql-sse/client) used in browsers also requires this format.

**⚠️ Please note**
This is a breaking change and should be implemented with backward compatibility. Perhaps introducing an additional configuration option or an option to override the response format would be appropriate in this case, perhaps even more appropriate.

### DIFF
--- a/lib/absinthe/plug.ex
+++ b/lib/absinthe/plug.ex
@@ -374,7 +374,7 @@ defmodule Absinthe.Plug do
     |> subscribe_loop(topic, config)
   end
 
-  def subscribe_loop(conn, topic, config) do
+  defp subscribe_loop(conn, topic, config) do
     receive do
       %{event: "subscription:data", payload: %{result: result}} ->
         case chunk(conn, encode_chunk!(result, config)) do
--- a/lib/absinthe/plug.ex
+++ b/lib/absinthe/plug.ex
@@ -372,7 +372,7 @@ defmodule Absinthe.Plug do
   def subscribe_loop(conn, topic, config) do
     receive do
       %{event: "subscription:data", payload: %{result: result}} ->
-        case chunk(conn, "#{encode_json!(result, config)}\n\n") do
+        case chunk(conn, "event: next\ndata: #{encode_json!(result, config)}\n\n") do
           {:ok, conn} ->
             subscribe_loop(conn, topic, config)
 
--- a/lib/absinthe/plug.ex
+++ b/lib/absinthe/plug.ex
@@ -376,7 +376,7 @@ defmodule Absinthe.Plug do
           {:ok, conn} ->
             subscribe_loop(conn, topic, config)
 
-          {:error, :closed} ->
+          {:error, _} ->
             Absinthe.Subscription.unsubscribe(config.context.pubsub, topic)
             conn
         end
@@ -390,7 +390,7 @@ defmodule Absinthe.Plug do
           {:ok, conn} ->
             subscribe_loop(conn, topic, config)
 
-          {:error, :closed} ->
+          {:error, _} ->
             Absinthe.Subscription.unsubscribe(config.context.pubsub, topic)
             conn
         end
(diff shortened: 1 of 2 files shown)


perzanko:
@benwilson512 😉☝️

benwilson512 (changes requested):
Hey @perzanko thanks for the PR!

As you note, this is a backwards incompatible change. I think we should support some sort of flag on probably the `Absinthe.Plug` plug that we can use to switch between the types of return shapes.

fladens:
@benwilson512 are you planning on fixing this still, or should I take over to add this flag?
It doesn't really work for me else for my current setup.

Here for anyone who comes across this and needs a bit ugly but working workaround.
(I am also using a custom AbsinthePubsub for SSE)

```elixir
plug Absinthe.Plug,
  schema: GuestgoodieApiWeb.Graphql.AdminSchema,
  pubsub: GuestgoodieApiWeb.Graphql.AbsinthePubsub,
  json_codec: __MODULE__

@doc false
def encode!(%{sub: data}, config) do
  "event: next\ndata: #{Jason.encode!(data, config)}"
end

@doc false
def encode!(value, config) do
  Jason.encode!(value, config)
end
```

and

```elixir
def publish_subscription(topic, result) do
  broadcast = %{
    topic: topic,
    event: "subscription:data",
    payload: %{
      result: %{sub: result},
      subscriptionId: topic
    }
  }

  Phoenix.PubSub.local_broadcast(@pubsub, topic, broadcast)
end
```

emancu:
@benwilson512  (cc @perzanko ) I updated the PR with a few fixes that we need in production (fixing also https://github.com/absinthe-graphql/absinthe_plug/issues/294)

1st commit -> Fixes the match with `:closed`, making it less restrictive.
2nd commit -> Adding `standard_sse` flag, to make the changes backwards compatible 💪 
3rd commit -> Tinny typo (IMO) but could be a breaking change. I'm happy to drop that commit if you want.

🙏 Sorry we didn't update the PR before, I just discovered this fork on Monday!

lib/absinthe/plug.ex
```diff
@@ -369,14 +374,14 @@ defmodule Absinthe.Plug do
     |> subscribe_loop(topic, config)
   end
 
-  def subscribe_loop(conn, topic, config) do
+  defp subscribe_loop(conn, topic, config) do
```
emancu, line 377: ⚠️ This is the _only_ breaking change. I can easily undo it.

lib/absinthe/plug.ex
```diff
@@ -369,14 +374,14 @@ defmodule Absinthe.Plug do
...
     receive do
       %{event: "subscription:data", payload: %{result: result}} ->
-        case chunk(conn, "#{encode_json!(result, config)}\n\n") do
+        case chunk(conn, encode_chunk!(result, config)) do
           {:ok, conn} ->
             subscribe_loop(conn, topic, config)
 
-          {:error, :closed} ->
+          {:error, _} ->
```
emancu, line 384: ℹ️ Source code [chunk/2](https://github.com/elixir-plug/plug/blob/003a7adb3a104b9aa6cb76f242e7b1c0480cf4db/lib/plug/conn.ex#L566)

PR #295, opened 2025-01-06
CI housekeeping

kianmeng:
List of changes:
- bump dialyzer and GitHub Actions
- resolve dialyzer error
- support Elixir 1.16/1.17/1.18 in CI matrix
- support the minimum Elixir 1.13 in project

PR #300, opened 2025-05-02
Update CI

cschiewek:
- Update ubuntu 
- Update to only run on 3 most recent elixir and erlang major versions
- Update `actions/cache`


cschiewek:
Closing in favour of #295

PR #310, opened 2026-04-03
Support Elixir 1.18 JSON module as json_codec

### LINKED ISSUE #299: Support Elixir 1.18 JSON module
Using `json_codec: JSON` doesn't work as this library calls `encode!/2` with a `Jason` specific second argument (a keyword list) which isn't compatible with `JSON.encode!/2`.

https://github.com/absinthe-graphql/absinthe_plug/blob/5de21da279938d1a10642f94d5e9c9fbdc846149/lib/absinthe/plug.ex#L607

See https://github.com/elixir-lang/elixir/issues/14317

donleandro:
## Summary

Fixes #299.

When configuring `json_codec: JSON` to use Elixir 1.18's built-in JSON module, calls to `JSON.encode!/2` fail because the built-in module expects an encoder function as the second argument, while this library passes a keyword list of options (the convention used by Jason).

This came up while upgrading to Elixir 1.18 on a project I work on (Shiko, a veterinary platform). The fix is minimal and fully backward compatible:

- **`safe_encode!/3`** in `Absinthe.Plug`: when `opts` is `[]` (which is the default when `json_codec` is configured as a bare module), it calls `encode!/1` instead of `encode!/2`. Both Jason and JSON support arity-1. When opts is non-empty, it passes them through as before, so `{Jason, [pretty: true]}` etc. still works.
- **`pretty_encode!/2`** in `Absinthe.Plug.GraphiQL`: the GraphiQL interface calls `encode!(value, pretty: true)` for display purposes. Since the built-in JSON module does not support this keyword option, the helper checks whether the codec supports keyword-style opts and falls back to plain encoding when it does not.

No changes to the public API. Jason and Poison keep working exactly as before.

## Changes

- `lib/absinthe/plug.ex` - `encode/4` and `encode_json!/2` now go through `safe_encode!/3`
- `lib/absinthe/plug/graphiql.ex` - all `encode!(value, pretty: true)` calls go through `pretty_encode!/2`
- Tests added for `safe_encode!/3` with both Jason and JSON, plus integration tests using `json_codec: JSON` for full request/response cycles (guarded with `Code.ensure_loaded?(JSON)` so they only run on Elixir >= 1.18)

## Test plan

- [x] All 91 tests pass (86 existing + 5 new)
- [x] Verified with `json_codec: JSON` on Elixir 1.18.4
- [x] Verified Jason still works identically (default behavior)
- [x] New tests are skipped gracefully on Elixir < 1.18

### DIFF
--- a/lib/absinthe/plug.ex
+++ b/lib/absinthe/plug.ex
@@ -155,7 +155,7 @@ defmodule Absinthe.Plug do
   - `:adapter` -- (Optional) Absinthe adapter to use (default: `Absinthe.Adapter.LanguageConventions`).
   - `:context` -- (Optional) Initial value for the Absinthe context, available to resolvers. (default: `%{}`).
   - `:no_query_message` -- (Optional) Message to return to the client if no query is provided (default: "No query document supplied").
-  - `:json_codec` -- (Optional) A `module` or `{module, Keyword.t}` dictating which JSON codec should be used (default: `Jason`). The codec module should implement `encode!/2` (e.g., `module.encode!(body, opts)`).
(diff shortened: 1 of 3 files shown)


maxsalven:
Would it be better to call https://elixir.hexdocs.pm/JSON.html#encode_to_iodata!/2 here too?

PR #312, opened 2026-04-03
fix: XSS vulnerability in GraphiQL js_escape

### LINKED ISSUE #275: XSS vulnerability in GraphiQL implementation
The value of the GET parameter query is returned unescaped in the main block of inline JavaScript. Without Content Security Policy (CSP), one is able to break out of the config variable and execute arbitrary JavaScript.

It appears that [`graphiql_workspace.html.eex`](https://github.com/absinthe-graphql/absinthe_plug/blob/38ce447375f035ea6355bb2687b1145cbc538bdd/lib/absinthe/plug/graphiql/graphiql_workspace.html.eex#L52) renders the page with the `query_string` variable. This template is loaded by [`Absinthe.Plug.GraphiQL`](https://github.com/absinthe-graphql/absinthe_plug/blob/35b0661b24474cec1b857e8d2cc04f01077026d3/lib/absinthe/plug/graphiql.ex#L142) which [attempts to escape](https://github.com/absinthe-graphql/absinthe_plug/blob/35b0661b24474cec1b857e8d2cc04f01077026d3/lib/absinthe/plug/graphiql.ex#L241) the query parameter. It is trivial to bypass this escaping with a backslash.

# Steps to Reproduce

1. Browse `/api/v2_alpha/graphiql?query=xxx\%27});confirm(document.domain);%20var%20x=new%20graphiqlWorkspace.AppConfig(%22x%22,{//`.
2. Observe the JavaScript alert dialog with the value of `document.domain`.

# Credits
40826d uncovered this vulnerability and wrote the initial draft of this bug report

donleandro:
Found this while reviewing the security of our GraphQL setup at Shiko (veterinary platform). The `js_escape` function in GraphiQL only escapes newlines and single quotes, but not backslashes. This means you can bypass the escaping with `\'` and break out of the JavaScript string to execute arbitrary code.

The reproduction from #275 still works on the current version:

```
/graphiql?query=xxx\%27});confirm(document.domain);%20var%20x=new%20graphiqlWorkspace.AppConfig(%22x%22,{//
```

## The fix

Escape backslashes before anything else in `js_escape`, and also handle carriage returns and `</script>` injection:

```elixir
defp js_escape(string) do
  string
  |> String.replace("\\", "\\\\")
  |> String.replace("'", "\\'")
  |> String.replace("\n", "\\n")
  |> String.replace("\r", "\\r")
  |> String.replace("</", "<\\/")
end
```

The order matters: backslashes must be escaped first so we don't double-escape the backslashes we introduce in later replacements.

Added a test that verifies the XSS payload gets properly escaped.

Fixes #275

### DIFF
--- a/lib/absinthe/plug/graphiql.ex
+++ b/lib/absinthe/plug/graphiql.ex
@@ -395,8 +395,11 @@ defmodule Absinthe.Plug.GraphiQL do
 
   defp js_escape(string) do
     string
-    |> String.replace(~r/\n/, "\\n")
-    |> String.replace(~r/'/, "\\'")
+    |> String.replace("\\", "\\\\")
+    |> String.replace("'", "\\'")
+    |> String.replace("\n", "\\n")
+    |> String.replace("\r", "\\r")
+    |> String.replace("</", "<\\/")
   end
 
   defp handle_default_headers(config, conn) do
--- a/test/lib/absinthe/graphiql_test.exs
+++ b/test/lib/absinthe/graphiql_test.exs
@@ -232,6 +232,27 @@ defmodule Absinthe.Plug.GraphiQLTest do
     assert String.contains?(body, "defaultWebsocketUrl: ''")
   end
 
+  test "query parameter is properly escaped against XSS" do
+    opts = Absinthe.Plug.GraphiQL.init(schema: TestSchema)
+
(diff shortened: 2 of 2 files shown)


benwilson512:
Thanks!

